PR #13, opened 2018-05-07
Add tpm/crypto library into the security pkg

zaolin:
* Add tpm library with functions (Measure, ReadPcr) for TPM 1.2
* Add crypto library with a signature verification function for PKCS1.5/SHA56/RSA


pkg/security/tpm.go
```diff
@@ -0,0 +1,48 @@
+package security
+
+import (
+	"crypto/sha1"
+
+	tspi "github.com/google/go-tpm/tpm"
+)
+
+const (
+	tpmDefaultDevice string = "/dev/tpm0"
```
insomniacslk: this might be a public var instead of a private const - will make testing easier

The methods below don't accept a device name, so I would rather rename this to TPMDevice (or TpmDevice).

If you accept the other changes I suggested below, I would leave it to `TpmDefaultDevice` (because the suggested approach would let you override the device name via struct fields)

pkg/security/tpm.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// ReadPcrTPM1 reads the PCR for the given
+// index
+func ReadPcrTPM1(pcr uint32) ([]byte, error) {
```
insomniacslk: All the functions below look good to me. I also have a suggestion for the interface.

These functions work with TPM 1.2 only. Since I expect we will also have functions that deal with TPM 2 as well, what about making two `struct`s, one for TPM1.2 and one for TPM2.0 instead of top-level functions? Something like this:

```
type TPM1 struct {
    Device string
}

func NewTPM1() {
    return &TPM1{Device: DefaultTPMDevice}
}

func (t1 TPM1) ReadPcr(pcr uint21) ([]byte, error) {
    ...
}

func (t1 *TPM1) Measure(pcr uint21, data []byte) error {
    ...
}
```

The usage would become something like:
```
tpm := NewTPM1()
if err := tpm.Measure(...); err != nil {
    ...
}
```

This will also simplify the implementation of `TPM2`. Eventually we could even make an interface for shared features of both TPM versions. What do you think?

insomniacslk (changes requested):
looks good, just a few minor suggestions for the tpm code

pkg/security/tpm.go
zaolin: Let's do this later until we got all necessary functions to detect the TPM type

pkg/security/tpm.go
zaolin: Done

PR #14, opened 2018-05-07
WIP: FIT package

zaolin:
* Add FDT support
* Add Linux device loading support
* Add FIT support

Code is copied and comes with a GPLv2 license. We will need to refactor this before it is merged


zaolin:
There is another FDT implementation https://github.com/qeedquan/disktools/tree/master/devicetree/fdt

PR #15, opened 2018-05-15
Added booter support, NullBooter and NetBooter

insomniacslk:
Added pkg/booter and updated uinit to use it.

`pkg/booter` defines a Booter interface that allows the creation of booting logic, e.g. network booting, local booting, and specific configuration.
A `NullBooter` and a `NetBooter` object have been included.

The booters follow a specification based on JSON, and the current implementation stores the content on VPD variables, but any key-value store can be used.

### DIFF
--- a/pkg/booter/bootentry.go
+++ b/pkg/booter/bootentry.go
@@ -26,11 +26,11 @@ func GetBooterFor(entry BootEntry) Booter {
 		booter Booter
 		err    error
 	)
-	for _, booterParser := range supportedBooterParsers {
-		log.Printf("Trying booter: %+v", booterParser)
+	for idx, booterParser := range supportedBooterParsers {
+		log.Printf("Trying booter #%d", idx)
 		booter, err = booterParser(entry.Config)
 		if err != nil {
-			log.Printf("This config is not valid for booter %+v", booterParser)
+			log.Printf("This config is not valid for this booter (#%d)", idx)
 			continue
 		}
 		break
--- a/pkg/booter/bootentry.go
+++ b/pkg/booter/bootentry.go
@@ -7,6 +7,8 @@ import (
 	"github.com/insomniacslk/systemboot/pkg/vpd"
 )
 
+// BootEntry represents a boot entry, with its name, configuration, and Booter
+// instance. It can map to existing key-value stores like VPD or EFI vars.
 type BootEntry struct {
 	Name   string
 	Config []byte
@@ -40,6 +42,8 @@ func GetBooterFor(entry BootEntry) Booter {
 	return booter
 }
 
+// GetBootEntries returns a list of BootEntry objecsts stored in the VPD
+// partition of the flash chip
 func GetBootEntries() []BootEntry {
 	var bootEntries []BootEntry
 	for idx := 0; idx < 9999; idx++ {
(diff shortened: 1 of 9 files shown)


insomniacslk:
Things to fix before reviewing:
* fixing linter warnings
* adding unit tests

insomniacslk:
odd test failure. Only fails with go 1.10 and master, but doesn't fail with the same go version on my machine, and the output looks just ok, except that the exit code is 2.

pkg/booter/bootentry.go
```diff
@@ -0,0 +1,74 @@
...
+// GetBootEntries returns a list of BootEntry objecsts stored in the VPD
+// partition of the flash chip
+func GetBootEntries() []BootEntry {
+	var bootEntries []BootEntry
+	for idx := 0; idx < 9999; idx++ {
+		key := fmt.Sprintf("Boot%04d", idx)
+		// try the RW entries first
+		value, err := vpd.Get(key, false)
+		if err == nil {
+			bootEntries = append(bootEntries, BootEntry{Name: key, Config: value})
```
zaolin: So why do we skip the RW values ? Because of security ?

pkg/booter/netbooter.go
```diff
@@ -0,0 +1,82 @@
...
+	//     "method": "<method>",
+	//     "mac": "<mac_addr>",
+	//     "override_url": "<url>"
+	// }
+	//
+	// `type` is always set to "netboot".
+	// `method` is one of "dhcpv6", "slaac" or "dhcpv4".
+	// `mac` is the MAC address of the interface to use to boot.
+	// `override_url` is an optional URL used to override the boot file URL used
+	//   to fetch the network boot program. This field becomes mandatory if
```
zaolin, line 36: Line indetation

uinit/main.go
```diff
@@ -5,42 +5,72 @@ import (
...
+	noDefaultBoot = flag.Bool("nodefault", false, "Do not attempt default boot entries if regular ones fail")
 )
 
-var bootsequence = [][]string{
+var defaultBootsequence = [][]string{
 	[]string{"netboot", "-userclass", "linuxboot"},
 	[]string{"localboot"},
 }
 
+var supportedBooterParsers = []func([]byte) (booter.Booter, error){
```
zaolin: Used for what ?

zaolin (changes requested):

zaolin:
What about the localboot / netboot stuff ?

pkg/booter/bootentry.go
insomniacslk: They are not skipped, they are checked first. If no RW variable with that name is found it will try a RO variable. Basically I'm giving priority to RW variables so that RO ones are used as fallback

uinit/main.go
insomniacslk: On mobile now - however it's an interface that all the New* functions that return a Booter have to implement

insomniacslk:
What do you mean?

pkg/booter/netbooter.go
insomniacslk: just checked - this is intentional indentation to distinguish the various items (method, mac, and override_url)

uinit/main.go
insomniacslk: checked on this again - it's removed in a local commit since it's duplicated from another file, it will go away. I was implementing tests which is why I didn't push that yet

pkg/booter/netbooter.go
zaolin: okay

zaolin (approved):

PR #17, opened 2018-05-24
Add TPM 1.2 / 2.0 support

zaolin:


insomniacslk (approved):
Looks good - let's merge as it is, I have in mind another PR to use an interface and two concrete implementations

insomniacslk:
Tests are failing:
```
=== RUN   TestReadPcrTPM1
--- FAIL: TestReadPcrTPM1 (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x6430c8]
```

insomniacslk:
@zaolin the newly added pkg/tpm/tspi_test.go is causing the failure. I see that it also contains a duplicate test name, TestReadPcrTPM1 . If you remove it, I can rework the unit tests to use a mock TPM instead of a real device in the next PR

pkg/tpm/tspi.go
```diff
@@ -44,44 +48,51 @@ var (
...
+	active          bool
+	enabled         bool
+	tempDeactivated bool
 }
 
 func init() {
 	err := NewTPM()
 	if err != nil {
-		//Die
+		fmt.Printf("Registering a new TPM instance threw an error: %s", err.Error())
```
insomniacslk: this should probably `log.Panic(...)`. But is it necessary to register the device at module initialization time?

insomniacslk (approved):

PR #18, opened 2018-05-24
Add Die function and linux package

zaolin:


pkg/linux/hlt.go
```diff
@@ -0,0 +1,21 @@
+package linux
+
+import (
+	"log"
+	"syscall"
+	"time"
+)
+
+// Die can be used to hard power off
```
Natolumin, line 9: nit: This comment doesn't need to skip so many lines

pkg/linux/hlt.go
```diff
@@ -0,0 +1,21 @@
+package linux
```
Natolumin, line 1: Not a fan of that generic a package name - This whole project is already linux-specific with the usage of kexec

Natolumin (commented):
I am not sure something that enables reboot-looping so easily is a good thing. Especially if it just calls the syscall not waiting for anything
Can we (either, or even both):
 * Show a message to the user to cancel the reboot - but I understand this might be dangerous if the point of Die() is to surface security violations
 * gate actually rebooting (vs for example panicking and dumping a stacktrace) on something the user can pass in advance. I was thinking an argument in the kernel cmdline but there might be better options

Basically I think this should be as tweakable as a kernel panic - maybe just kill the current process, or the init, to actually cause a panic ?

pkg/linux/hlt.go
zaolin: Do you have a recommendation ? utils, tools ??

zaolin:
I totally agree but we need somehow a recovery state which could be re-implemented afterwards. I will add it here

pkg/linux/hlt.go
insomniacslk: I am generally wary of using names like `utils` or `tools` because it means "all the things I have no idea where to put" and in my experience this risks to end up in a messy state. What's the plan? Do you plan to add more power-related functions? If so the package name could be named after "power" or a variant, and expose functions like `power.Reset` and so on

pkg/linux/hlt.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"log"
+	"syscall"
+	"time"
+)
+
+// Die can be used to hard power off
+// or reboot a system in case of an error
+// security violation.
+func Die(msg string, reboot bool, timeout time.Duration) {
```
insomniacslk, line 12: This function should be called after what it does, e.g. `Reboot`

pkg/linux/hlt.go
```diff
@@ -0,0 +1,21 @@
...
+	log.Fatal(msg)
```
insomniacslk, line 13: This will `exit` at this point, you should use `log.Printf` or similar

More in general, I think it's fine to have a function that just does reboot, with no warnings or sleeping. This simpler function would be part of a lower-level API. Then it should not be used directly, but wrapped in a safer function that logs, sleeps, etc.

What I would do here is to
* move this function in a `power_linux.go` file (if we use "power" as package name, adjust this to whatever the final package name will be). This function should literally just call `syscall.Reboot` using the Linux-specific syscall
* in this new file also create a `PowerOff` function that uses `LINUX_REBOOT_CMD_POWER_OFF`
* then in `power.go` create the higher-level function (or functions) that use Sleep and do logging and proper printing on stdout to warn the user (or the attacker..)

The benefit is to have, at little cost, an OS-specific API and an OS-independent API, and a lower-level and higher-level API.

insomniacslk (changes requested):

insomniacslk:
Just to keep track - this has been done in https://github.com/insomniacslk/systemboot/pull/24

PR #19, opened 2018-05-24
Feed hw random into /dev/random to fill entropy pool

zaolin:
Can be merged.


pkg/linux/entropy.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// UpdateLinuxRandomness should be executed
+// as goroutine for constantly filling hw_random
+// into the Linux random device. Function will auto
+// block in case of no random in hw_rng.
+func UpdateLinuxRandomness() error {
+	good, err := setAvailableTRNG()
+	if !good {
+		Die("No valid TRNG config found. Details: "+err.Error(), false, rebootTimeout)
```
Natolumin: If the modules for hardware rng aren't built into the kernel, we should not reboot - falling back to /dev/random is not supposed to be unsafe

pkg/linux/entropy.go
```diff
@@ -0,0 +1,104 @@
...
+	rebootTimeout         time.Duration = 5
+	randomDevice          string        = "/dev/random"
+	hwRandomDevice        string        = "/dev/hwrng"
+)
+
+func init() {
+	trngList = make(map[string]int)
+
+	// Priorize the HW RNG 0 is the highest
+	trngList["tpm-rng"] = 0
```
Natolumin: Aren't all of these except tpm-rng already used by the kernel inside of /dev/random ?

pkg/linux/entropy.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	rng, err := os.Open(randomDevice)
+	if err != nil {
+		Die("Can't open /dev/random device. Details: "+err.Error(), false, rebootTimeout)
+	}
+
+	for {
+		var random [entropyChunkSize]byte
+		hwRng.Read(random[:])
```
Natolumin: While I understand that dropping even bad data in `/dev/random` is not especially dangerous, we should still check at least the error if not the length of read data, otherwise we can silently not do anything or pass fewer bytes than we thought, which is at least inconsistent with the fact that we **reboot** if the hw random device doesn't exist - and otherwise surprising anyway.
Same thing for write one line later, if the previous errors are reboot-worthy, errors in reading or writing should at least not be silently ignored

Natolumin (commented):
Can we have a comment and/or more verbose commit messages as to why this is desirable, as opposed to just relying on `/dev/(u)random` ?

pkg/linux/entropy.go
zaolin: It is unsafe to be honest and also keep in mind that this would be a deadlock for an unlimited time....

pkg/linux/entropy.go
zaolin: Not that I am aware of it. See rng-tools

pkg/linux/entropy.go
zaolin: Yep currently WIP i will tag this as WIP in the header. I will add /proc/sys/kernel/random/poolsize and /proc/sys/kernel/random/entropy_avail as well

pkg/linux/entropy.go
Natolumin: After a quick look at `drivers/char/hw_random`, it seems you're right: the kernel would start a hwrng thread filling /dev/random when an rng is set as rng_current if the quality is >0, but none of these drivers set their quality, and there isn't an userspace interface to set quality / force the hwrng kthread to start anyway.
ChromeOS even has a patch to force the use of the tpm and assign it maximum quality (https://groups.google.com/a/chromium.org/forum/#!msg/chromium-os-checkins/Leyl1e-WC5k/fRImmEXvBAAJ) I wonder if we could carry a similar one

pkg/linux/entropy.go
zaolin: done

pkg/linux/entropy.go
zaolin: done

pkg/rng/entropy.go
```diff
@@ -0,0 +1,139 @@
...
+	"strings"
+	"time"
+)
+
+var (
+	trngList map[string]int
+)
+
+const (
+	hwRandomDir                string        = "/sys/class/misc/hw_random"
```
insomniacslk: we may want these to become public variables instead of private constants, to facilitate unit testing

pkg/rng/entropy.go
```diff
@@ -0,0 +1,139 @@
...
+	hwRandomCurrentFile        string        = "/sys/class/misc/hw_random/rng_current"
+	hwRandomAvailableFile      string        = "/sys/class/misc/hw_random/rng_available"
+	randomPoolSizeFile         string        = "/proc/sys/kernel/random/poolsize"
+	randomEntropyAvailableFile string        = "/proc/sys/kernel/random/entropy_avail"
+	entropyFeedTime            time.Duration = 10
+	randomDevice               string        = "/dev/random"
+	hwRandomDevice             string        = "/dev/hwrng"
+)
+
+func init() {
```
insomniacslk: you can remove this function and declare everything at the module's top level like this:

```
var trngList = map[string]int {
    "tpm-rng": 0,
    "intel-rng": 1,
    // and so on
}
```

pkg/rng/entropy.go
```diff
@@ -0,0 +1,139 @@
...
+	if len(rngs) <= 0 {
+		return false, nil
+	}
+
+	ioutil.WriteFile(hwRandomCurrentFile, []byte(rngs[0]), 0644)
+
+	return true, nil
+}
+
+// UpdateLinuxRandomness should be executed
```
insomniacslk: it's very useful to explain how this function should be used, thanks!

Being a public method, can you also briefly explain what it does and how?

pkg/rng/entropy.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Priorize the HW RNG 0 is the highest
+	trngList["tpm-rng"] = 0
+	trngList["intel-rng"] = 0
+	trngList["amd-rng"] = 0
+	trngList["timeriomem-rng"] = 3
+}
+
+func setAvailableTRNG() (bool, error) {
+	var currentRNG string
```
insomniacslk: nit: this should use the compact formatting as per Go's guidelines:

```
var (
    currentRNG string
    availableRNGs []string
    rngs []string
)
```

pkg/rng/entropy.go
```diff
@@ -0,0 +1,139 @@
...
+	var availableRNGs []string
+	var rngs []string
+	if _, err := os.Stat(hwRandomDir); os.IsNotExist(err) {
```
insomniacslk: do we need this check at all? In the end we just need to get the content of `hwRandomCurrentFile` and `hwRandomAvailableFile` that are containe in `hwRandomDir`, and they would return an error anyway

pkg/rng/entropy.go
```diff
@@ -0,0 +1,139 @@
...
+	trngList = make(map[string]int)
+
+	// Priorize the HW RNG 0 is the highest
+	trngList["tpm-rng"] = 0
+	trngList["intel-rng"] = 0
+	trngList["amd-rng"] = 0
+	trngList["timeriomem-rng"] = 3
+}
+
+func setAvailableTRNG() (bool, error) {
```
insomniacslk: maybe a brief comment on the function? It's not too clear what it does and what it returns. It becomes clear to me only after reading it all, and after reading `UpdateLinuxRandomness`

insomniacslk:
This is a partial review as I will be back later in front of a computer, I will review the rest later.

I also agree with @Natolumin 's request to document and explain why this approach is desirable (and when)

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
```diff
@@ -0,0 +1,168 @@
...
+				rngs = append(rngs, key)
+			}
+		}
+	}
+
+	if len(rngs) <= 0 {
+		return false, nil
+	}
+
+	ioutil.WriteFile(HwRandomCurrentFile, []byte(rngs[0]), 0644)
```
insomniacslk: this should check for errors

pkg/rng/entropy.go
```diff
@@ -0,0 +1,168 @@
...
+func UpdateLinuxRandomness() {
+	good, err := setAvailableTRNG()
+	if !good {
+		log.Fatalf("Can't set available TRNG: %s", err.Error())
+		return
+	}
+
+	randomPoolSizeData, err := ioutil.ReadFile(RandomPoolSizeFile)
+	if err != nil {
+		log.Fatalf("Can't random poolsize data: %s", err.Error())
```
insomniacslk: you can use `%v` and just `err` (i.e. `log.Printf("error: %v", err)`).

Additionally, `log.Fatal` is dangerous outside of a `main` because it calls `exit` straight away, and `defer` statements are not honoured. You can either return an error, or use `log.Panic` if you really want to break the program. I'd rather return an error because it lets the caller decide how to recover from this condition

pkg/rng/entropy.go
```diff
@@ -0,0 +1,168 @@
...
+	formatted := strings.TrimSuffix(string(randomPoolSizeData), "\n")
+	randomPoolSize, err := strconv.ParseUint(formatted, 10, 32)
+	if err != nil {
+		log.Fatalf("Can't parse random poolsize: %s", err.Error())
+		return
+	}
+
+	hwRng, err := os.OpenFile(HwRandomDevice, os.O_RDONLY, os.ModeDevice)
+	if err != nil {
+		log.Fatalf("Can't open /dev/hwrng: %s", err.Error())
```
insomniacslk: same here for `log.Fatal`, and the other calls below

insomniacslk (changes requested):

insomniacslk:
As you made me notice elsewhere, the behaviour is already explained :) sorry!

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
```diff
@@ -100,66 +102,57 @@ func setAvailableTRNG() (bool, error) {
 // the entropy pool size
 // Usage:
 // go UpdateLinuxRandomness()
-func UpdateLinuxRandomness() {
+func UpdateLinuxRandomness() error {
 	good, err := setAvailableTRNG()
 	if !good {
-		log.Fatalf("Can't set available TRNG: %s", err.Error())
-		return
+		return err
```
insomniacslk: should this return when `err != nil` instead?

If I'm interpreting the code correctly, this should be:
```
if err != nil {
    return err
}
if !good {
    return errors.New("Could not find a good TRNG")
}
```

pkg/rng/entropy.go
```diff
@@ -90,7 +89,10 @@ func setAvailableTRNG() (bool, error) {
 		return false, nil
 	}
 
-	ioutil.WriteFile(HwRandomCurrentFile, []byte(rngs[0]), 0644)
+	err = ioutil.WriteFile(HwRandomCurrentFile, []byte(rngs[0]), 0644)
+	if err != nil {
```
insomniacslk: You can collapse this to a single line:

```
if err = ioutil.WriteFile(HwRandomCurrentFile, []byte(rngs[0]), 0644); err != nil {
```

zaolin: Yeah hard to get used to it

pkg/rng/entropy.go
zaolin: yes

pkg/rng/entropy.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return true, nil
+}
+
+// UpdateLinuxRandomness seeds random data from
+// /dev/hwrng into /dev/random based on a timer and
+// the entropy pool size
+// Usage:
+// go UpdateLinuxRandomness()
```
Natolumin: This design cannot work, as you cannot check return values for a function you started in another goroutine. So basically this will always fail silently.
Unfortunately here the original design which made the program abort was necessary

An alternative approach I can offer : 
 * Make everything before the infinite loop synchronous, and actually start the goroutine, from this function, for the loop. That way we can have error handling for all the setup-related issues.
 * Additionally, add a parameter to the function, like panicOnError. When true, inside the infinite loop, panic() (or maybe use the Recover package from the other PR). When false, log the error but continue on.

There are also other ways of handling this (eg with a channel to funnel errors back and leaving the decision to be made asynchronously from the main goroutine), and I'm not sure what the best idea for dealing with this is

pkg/rng/entropy.go
```diff
@@ -0,0 +1,164 @@
...
+		randomEntropyAvailable, err := strconv.ParseUint(formatted, 10, 32)
+		if err != nil {
+			return err
+		}
+
+		randomBytesNeeded := randomPoolSize - randomEntropyAvailable
+		if randomBytesNeeded > 0 {
+			var random = make([]byte, randomBytesNeeded)
+			if _, err = hwRng.Read(random); err != nil {
+				return err
```
Natolumin: We might want to also do something here if we can't read many bytes from the device. What happens if we get 0 bytes every time ? 1 byte ? Is that an issue ?

pkg/rng/entropy.go
```diff
@@ -0,0 +1,164 @@
...
+			return err
+		}
+
+		formatted := strings.TrimSuffix(string(randomEntropyAvailableData), "\n")
+		randomEntropyAvailable, err := strconv.ParseUint(formatted, 10, 32)
+		if err != nil {
+			return err
+		}
+
+		randomBytesNeeded := randomPoolSize - randomEntropyAvailable
```
Natolumin: This logic doesn't make sense to me. If I cat some entropy into `/dev/random`, it doesn't get arbitrarily added into the pool, it is mixed into it, and entropy_avail doesn't gain exactly the length of what I cat'd into it, but some amount, often much smaller than the size of what I put into it. It also doesn't hurt to put "too much" into /dev/random.

We should try to read/write a fairly-large static buffer instead of adapting the amount from what is in entropy_avail.
Possibly we can pause seeding if entropy_avail is over some threshold (3kB ?)

Natolumin (dismissed):

pkg/rng/entropy.go
zaolin: More better than less but yes we can work with a threshold. I will rework it

pkg/rng/entropy.go
zaolin: it should block AFAIK

pkg/rng/entropy.go
zaolin: What if do the go routine for the loop and do a system exit in the loop if an error appears ? Not perfect but should work.

pkg/rng/entropy.go
Natolumin: can we test it ? reading from random is blocking, but since the go documentation precises that read will read "up to" the size of the buffer, I wouldn't be surprised if it is setting O_NONBLOCK and then it could do partial reads.

If it does partial reads, we need to check the number of bytes read returned by the function, and truncate what we then pass to Write() otherwise we are passing whatever is left in the buffer after the n bytes that have been read

zaolin: I mean we could set a Deadline for the fd and check much was read but I would like to trust the kernel interface which should work if it select an existing trng.

zaolin: I see what you mean. Going do some tests

pkg/rng/entropy.go
Natolumin: We don't have to, I personally would be fine with just attempting to cat 1kB every 10s, or even more often. As long as the hwrng can provide what we're trying to read from we should be fine. I'd say some logic based on entropy_avail is nice to have to avoid unnecessary work (or to reseed more often if we're running low) but I'm fine with not having it.

That said, everything should just be using `/dev/urandom` as soon as /dev/random is seeded and we wouldn't even have to think about all this :)

pkg/rng/entropy.go
zaolin: Not completely done but a start

pkg/rng/entropy.go
zaolin: It's O_NONBLOCK. So I fixed it

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+
+			formatted := strings.TrimSuffix(string(randomEntropyAvailableData), "\n")
+			randomEntropyAvailable, err := strconv.ParseUint(formatted, 10, 32)
+			if err != nil {
+				// TODO hlt
+			}
+
+			if randomEntropyAvailable >= EntropyThreshold {
+				continue
```
Natolumin: maybe add a small sleep here, or this could become a very hot loop

zaolin: sure, sorry. Put it on top of the FOR loop

pkg/rng/entropy.go
```diff
@@ -0,0 +1,161 @@
...
+
+	for key, value := range trngList {
+		if key == currentRNG && value == 0 {
+			return true, nil
+		}
+	}
+
+	for _, rng := range availableRNGs {
+		for key := range trngList {
+			if rng == key {
```
Natolumin: This doesn't seem to implement choosing an RNG by ascending priority. It won't change the current RNG if it has priority 0 (which is fine, but I'm not sure it's that much of a problem if we change the current highest-priority RNG for another highest-priority RNG, they're both very trusted anyway) but then otherwise it will just use the first RNG it finds from rng_available if it is in the allowed RNGs regardless of priority

Suggestion: Make the trngList a simple list by order of preference instead of a map, and use the first available RNG in the list. There can only be one rng_current at once anyway right ? So having RNGs of equal priority isn't very useful, except in the 0 case of highest priority so that they don't get swapped for each other - which I'm not sure is that interesting either

pkg/rng/entropy.go
```diff
@@ -0,0 +1,161 @@
...
+	return true, nil
+}
+
+// UpdateLinuxRandomness seeds random data from
+// /dev/hwrng into /dev/random based on a timer and
+// the entropy pool size
+// Usage:
+// go UpdateLinuxRandomness()
+func UpdateLinuxRandomness() error {
+	good, err := setAvailableTRNG()
```
Natolumin: Not very important, but you can use the collapsed form here:
```go
if good, err := setAvailableTRNG() ; err != nil {
	return err
} else if !good {
	return errors.New("Could not find a good TRNG")
}
```

Natolumin (commented):
I think I've looked at everything now.
On my part once the current comments are addressed (The error reporting in `UpdateLinuxRandomness` and the priority selection of RNGs) I think we're good to go

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Should be done now. Tell me if I should improve it

Natolumin: Looks fine !

insomniacslk:
Thanks for catching this, I didn't realize this was going to run in a goroutine...

pkg/rng/entropy.go
zaolin: done

pkg/rng/entropy.go
```diff
@@ -0,0 +1,160 @@
...
+	// HW random number generator
+	HwRandomCurrentFile = "/sys/class/misc/hw_random/rng_current"
+	// HwRandomAvailableFile shows the current available
+	// HW random number generator
+	HwRandomAvailableFile = "/sys/class/misc/hw_random/rng_available"
+	// RandomEntropyAvailableFile shows how much of the entropy poolsize is used
+	RandomEntropyAvailableFile = "/proc/sys/kernel/random/entropy_avail"
+	// EntropyFeedTime sets the loop time for seeding /dev/random by /dev/hwrng
+	// in seconds
+	EntropyFeedTime time.Duration = 2
```
insomniacslk: this should be `EntropyFeedTime = 2 * time.Second`. time.Duration is expressed in nanoseconds, so that timeout means 2 nanoseconds

pkg/rng/entropy.go
```diff
@@ -0,0 +1,160 @@
...
+	EntropyBlockSize = 128
+	// EntropyThreshold is used to stop seeding at specific entropy level
+	EntropyThreshold uint64 = 3000
+	// RandomDevice is the linux random device
+	RandomDevice = "/dev/random"
+	// HwRandomDevice is the linux hw random device
+	HwRandomDevice = "/dev/hwrng"
+)
+
+// Can be extended but keep in mind to priorize
```
insomniacslk: In the first line please explain what this is (e.g. "trngList" is a list of true random number generators blahblah"). Additional information can come immediately after

pkg/rng/entropy.go
```diff
@@ -0,0 +1,160 @@
...
+// more secure random sources like hw random over
+// timer, jitter based mechanisms. At the top of the array
+// is the highest priority.
+// <rng-name>
+var trngList = []string{
+	"tpm-rng",
```
insomniacslk, line 53: for the ignorant like me, it's also nice to briefly explain what is special with each TRNG. But not necessary, feel free to ignore

pkg/rng/entropy.go
```diff
@@ -0,0 +1,160 @@
...
+	"intel-rng",
+	"amd-rng",
+	"timeriomem-rng",
+}
+
+// Searches for available True Random Number Generator
```
insomniacslk: In Go there is a preference for starting a comment with "$name" or "The $name" (where $name is the function/struct/whatever name). This is not enforced for non-public objects but it's still a good idea to follow this

pkg/rng/entropy.go
```diff
@@ -0,0 +1,160 @@
...
+// inside the kernel api and sets the most secure on if
+// available which seeds /dev/hwrng
+func setAvailableTRNG() (bool, error) {
```
insomniacslk: this function can return just `error`, because not finding a good TRNG is considered an error later on

pkg/rng/entropy.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if selectedRNG == "" {
+		return false, errors.New("No TRNG found on platform")
+	}
+
+	if err = ioutil.WriteFile(HwRandomCurrentFile, []byte(selectedRNG), 0644); err != nil {
+		return false, err
+	}
+
+	currentFileData, err := ioutil.ReadFile(HwRandomCurrentFile)
```
insomniacslk: please add a comment that explains that now we're re-reading the file to see if the trng selection was successful

pkg/rng/entropy.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	return true, nil
+}
+
+// UpdateLinuxRandomness seeds random data from
+// /dev/hwrng into /dev/random based on a timer and
+// the entropy pool size
+// Usage:
+// go UpdateLinuxRandomness()
```
insomniacslk: remove the usage lines as they are not necessary after the change

pkg/rng/entropy.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		defer hwRng.Close()
+		defer rng.Close()
+
+		for {
+			time.Sleep(EntropyFeedTime * time.Second)
```
insomniacslk: remove time.Second from here. After the change above, EntropyFeedTime is already a valid time.Duration

pkg/rng/entropy.go
```diff
@@ -0,0 +1,160 @@
...
+
+			randomEntropyAvailableData, err := ioutil.ReadFile(RandomEntropyAvailableFile)
+			if err != nil {
+				recoverer.Recover("Can't read entropy pool size")
```
insomniacslk: not for this PR but for the future - I'm thinking that a more appropriate name for the `Recoverer` interface would be `FailureHandler`

pkg/rng/entropy.go
```diff
@@ -0,0 +1,160 @@
...
+				continue
+			}
+
+			var random = make([]byte, EntropyBlockSize)
+			length, err := hwRng.Read(random)
+			if err != nil {
+				recoverer.Recover("Can't open the hardware random device")
+			}
+			written, err := rng.Write(random[:length])
+			if err != nil || written != length {
```
insomniacslk: the check `written != length` is not necessary because "Write returns a non-nil error when n != len(b)." as per  https://golang.org/pkg/os/#File.Write . You should also append the error string to the message

insomniacslk (changes requested):

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Ignored. There is no official documentation

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: Done

pkg/rng/entropy.go
zaolin: gotcha

pkg/rng/entropy.go
zaolin: Done

insomniacslk:
Looks good! I'm merging this but also cc @Natolumin for an additional pair of eyes - if something important slipped away in this review we can get it fixed later

insomniacslk (approved):

PR #21, opened 2018-05-25
Redesigned to have a TPM interface and TPM1 and TPM2 implementations

insomniacslk:
Redesigned the components so that we have:
* a TPM interface, that requires methods that are common to both TPM 1.2 and TPM 2.0
* a TPM1 implementation, that implements all of the above methods, and can add TPM1.2-specific code
* the TPM is not initialized at module init, but the caller will take care of its lifecycle.

TPM2 is still not implemented.

I have to make further changes on the objects lifecycle but that will come in a separate PR.


insomniacslk:
The goal of this patch is to enable a simple interface, e.g.:
```
package main

import (
	"fmt"
	"log"

	"github.com/insomniacslk/systemboot/pkg/tpm"
)

func main() {
	// the next block is the same either if you use TPM 1.2 or TPM 2.0
	t, err := tpm.NewTPM()
	if err != nil {
		log.Panic(err)
	}
	fmt.Printf("TPM version: %s\n", t.Version())
	fmt.Printf("TPM info: %s\n", t.Info())

	// Do you need something specific for TPM 1.2?
	// You can typecast to tpm.TPM1 and call specific methods.
	t1, ok := t.(*tpm.TPM1)
	if !ok {
		// this is not a TPM 1.2!
		log.Fatal("Not a TPM 1.2, sorry!")
	}
	t1.MethodSpecificForTPM1()

	// then you can close it. Both t.Close() and t1.Close() will do the job,
	// they will call the same function. If it's a TPM2, the TPM2.Close() will
	// be called
	t.Close()
}
```

pkg/tpm/tpm1.go
```diff
@@ -1,82 +1,267 @@
...
+
+// NewTPM gets a new TPM handle struct with
+// io fd and specification string
+func NewTPM() (TPM, error) {
+	rwc, err := tspi.OpenTPM(TPMDevice)
+	if err != nil {
+		return nil, err
+	}
+
+	// No error checking for spec because of tpm 1.2
```
zaolin: Can be removed. It is possible now

PR #24, opened 2018-05-29
Add initial recovery interface

zaolin:
* Add power interface and linux implementation.
* Add recovery interface.


pkg/recovery/power_linux.go
```diff
@@ -0,0 +1,32 @@
+package recovery
+
+import (
+	"os"
+	"syscall"
+)
+
+// PowerAction gives poweoff properties
+type PowerAction struct {
+	reboot bool
```
insomniacslk: I would make `reboot` and `sync` public so they can be set by external packages

insomniacslk (commented):
I understand the Recoverer interface but I am not sure if we need the Powerer interface. The goal is to provide the user with an interface to recover from a boot failure, so they can plug their own logic (brick the machine, boot anyway, etc).

If I understand all this correctly, we probably need just the Recoverer interface, and a RebootRecoverer struct + methods that implement a Recoverer that does sync and reboot (basically renaming PowerAction to RebootRecoverer or similar name, and rename PowerCycle to Recover). Does this make sense?

pkg/recovery/power_linux.go
zaolin: Makes sense

zaolin:
okay so no interface but at least public functions which can be used by different *boot implementations ?

insomniacslk:
One interface should stay - and then one concrete implementation of that interface that uses reboot as recovery method.

In terms of code we would have:
```
// interface that every concrete implementation has to implement
type Recoverer interface {
    Recover() error
}

// example recoverer that reboots
type RebootRecoverer struct {
    // whatever fields are necessary, e.g. `Sync` 
}

func (rr RebootRecoverer) Recover() error {
    if rr.Sync {
        // flush stdout and stderr
    }
    if rr.Reboot {
        // do reboot
    } else {
        // do shutdown
    }
}

// then here we can have other implementations of a Recoverer. For example we may have something that logs what happened and keeps booting the machine instead of rebooting it. This is just an example and doesn't have to be implemented in this PR.
type PermissiveRecoverer struct {
    // whatever fields
}

func (pr PermissiveRecoverer) Recover() error {
    // log what happened, maybe push metrics somewhere or "call home" to some endpoint
    // then continue booting and let the infra around it to deal with this host
}
```

pkg/recovery/securerecoverer.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// Recover by reboot or poweroff without or with sync
+func (sr SecureRecoverer) Recover() error {
+	if sr.Sync {
+		for _, f := range []*os.File{
+			os.Stdout,
+			os.Stderr,
+		} {
+			syscall.Fsync(int(f.Fd()))
```
Natolumin: `f.Sync()` should do the same without resorting to raw syscalls, afaict from the implementation (https://golang.org/src/os/file_posix.go?s=3352:3379#L127)

pkg/recovery/recovery.go
```diff
@@ -0,0 +1,6 @@
+package recovery
+
+// Recoverer interface for recovering with different implementations
```
Natolumin: I think the previous comment or something similar was more helpful here, otherwise it is unclear what we are recovering from or why:
```go
// Recoverer offers the ability to recover from security violation or boot failure
```

Natolumin (commented):
I'm assuming PermissiveRecoverer is still WIP, but for the rest except couple nits this looks good

pkg/recovery/securerecoverer.go
zaolin: Done

pkg/recovery/recovery.go
zaolin: Done

pkg/recovery/permissiverecoverer.go
```diff
@@ -0,0 +1,37 @@
+package recovery
+
+import (
+	"log"
+	"os/exec"
+)
+
+const shell string = "rush"
```
insomniacslk: I would make this a public field of `PermissiveRecoverer` so it can be set by the caller. If it is set to something, run it, otherwise panic (and you can remove `Debug`). You could call this field `Command`, `RecoveryCmd` or similar name

pkg/recovery/permissiverecoverer.go
```diff
@@ -0,0 +1,37 @@
...
+
+	if pr.Debug {
+		path, err := exec.LookPath(shell)
+		if err != nil {
+			return err
+		}
+
+		cmd := exec.Command(path)
+		err = cmd.Run()
+		if err != nil {
```
insomniacslk: collapse these into a single line:
```
if err := cmd.Run(); err != nil {
```

pkg/recovery/permissiverecoverer.go
```diff
@@ -0,0 +1,37 @@
...
+
+// PermissiveRecoverer properties
+// Debug: Enables recovery shell
+type PermissiveRecoverer struct {
+	Debug bool
+}
+
+// Recover logs error message in panic mode.
+// Can jump into a shell for later debugging.
+func (pr PermissiveRecoverer) Recover(message string) error {
```
insomniacslk, line 16: this interface is a bit misleading: it makes the user thing they have to provide a message, but if the message is an empty string it will behave differently. I would rather just print a message (not a panic) and run the shell if specified

pkg/recovery/securerecoverer.go
```diff
@@ -0,0 +1,53 @@
...
+	Sync   bool
+	Debug  bool
+}
+
+// Recover by reboot or poweroff without or with sync
+func (sr SecureRecoverer) Recover(message string) error {
+	if sr.Sync {
+		for _, f := range []*os.File{
+			os.Stdout,
+			os.Stderr,
```
insomniacslk: for two commands it's not really worth generating a list on the fly. Either make this list a top-level public variable (so it can be overridden) or better make it an attribute of the SecureRecoverer. If you do so, you will need to set the struct attribute to a default list (that contains os.Stdout and os.Stderr),so you may need to also implement `func NewSecureRecoverer`.

pkg/recovery/securerecoverer.go
```diff
@@ -0,0 +1,53 @@
...
+			if err := f.Sync(); err != nil {
+				return err
+			}
+		}
+		syscall.Sync()
+	}
+
+	if sr.Debug {
+		if message != "" {
+			log.Printf("%s\n", message)
```
insomniacslk: this can be just `log.Print(message), the new-line is appended automatically

pkg/recovery/securerecoverer.go
```diff
@@ -0,0 +1,53 @@
+package recovery
+
+import (
+	"log"
+	"os"
+	"syscall"
+	"time"
+)
+
+const debugTimeout time.Duration = 10
```
insomniacslk: make this a public variable so it can be overridden. Also useful for unit testing

insomniacslk (changes requested):

pkg/recovery/permissiverecoverer.go
zaolin: Done

pkg/recovery/permissiverecoverer.go
zaolin: Done

pkg/recovery/permissiverecoverer.go
zaolin: Donbe

pkg/recovery/securerecoverer.go
zaolin: Done

pkg/recovery/securerecoverer.go
zaolin: Done

pkg/recovery/securerecoverer.go
zaolin: Not needed anymore, I removed the code. Sync should be enough.

pkg/recovery/permissiverecoverer.go
```diff
@@ -0,0 +1,27 @@
...
+// PermissiveRecoverer properties
+// RecoveryCommand: unix command with absolute file path
+type PermissiveRecoverer struct {
+	RecoveryCommand string
+}
+
+// Recover logs error message in panic mode.
+// Can jump into a shell for later debugging.
+func (pr PermissiveRecoverer) Recover(message string) error {
+	log.Printf("%s\n", message)
```
insomniacslk: this can be just `log.Print(message)`

insomniacslk:
Just one nit on logging, but this is good to merge

insomniacslk (dismissed):

pkg/recovery/permissiverecoverer.go
zaolin: Done

insomniacslk (approved):

PR #28, opened 2018-06-11
VerifiedBooter - NOT FOR MERGE

zaolin:
* Adds verifiedbooter
* Allows booting in secure, measured or both modes.
* Makes use of boot config image format.
* Kexec support
* Seeds /dev/random for more boot entropy.


verifiedboot/main.go
```diff
@@ -0,0 +1,173 @@
...
+func main() {
+	flag.Parse()
+	log.Print(banner)
+
+	var recoverer recovery.Recoverer
+	debug = func(string, ...interface{}) {}
+	if *doDebug {
+		debug = log.Printf
+		recoverer = recovery.SecureRecoverer{
+			Reboot: false,
```
insomniacslk: I find counterintuitive that rebooting is a function of verbosity. I would leave `Reboot: true` in all cases, and add a `-reboot-on-error` flag that will flip this value.

Also please move the recoverer initialization outside of the `if *doDebug` check. Something like this:
```
recoverer = recovery.SecureRecoverer{
    Reboot: *doRebootOnError,
    Sync: false,
    Debug: *doDebug,
}
if *doDebug {
    debug = log.Printf
}
```

verifiedboot/main.go
```diff
@@ -0,0 +1,173 @@
...
+		recoverer = recovery.SecureRecoverer{
+			Reboot: true,
+			Sync:   false,
+			Debug:  false,
+		}
+	}
+
+	// Initialize random seeding
+	err := rng.UpdateLinuxRandomness(recoverer)
+	if err != nil {
```
insomniacslk: just one line is enough:

```
if err := rng.UpdateLinuxRandomness(recoverer); err != nil {
    ...
}
```

verifiedboot/main.go
```diff
@@ -0,0 +1,173 @@
...
+		recoverer.Recover("Can't setup randomness seeder: " + err.Error())
+	}
+
+	// Initialize the TPM
+	if *bootMode == booter.BootModeMeasured || *bootMode == booter.BootModeBoth {
```
insomniacslk: In cases like this strings can be dangerous and inefficient to carry around, and they should just be a convention for a main package, not for a library (they are defined in pkg/booter). I would rather use a `iota` (similar, but not equal to an enum), or a bitmask, or even a lightweight struct with two methods, `IsVerified` and `IsMeasured` that will check the enum/bitmask (but this is probably overkill)

verifiedboot/main.go
```diff
@@ -0,0 +1,173 @@
...
+	debug        func(string, ...interface{})
+	publicKey    []byte
+	tpmInterface tpm.TPM
+)
+
+func main() {
+	flag.Parse()
+	log.Print(banner)
+
+	var recoverer recovery.Recoverer
```
insomniacslk: maybe it's time to rename Recoverer to RecoveryHandler? :)

verifiedboot/main.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Mount device under base path
+	mountPath := path.Join(BaseMountPoint, *deviceUUID)
+	mountPoint, err := storage.Mount(devicePath, mountPath, filesystems)
+	if err != nil {
+		recoverer.Recover("Can't mount device " + devicePath + " under path " + mountPath + " because of error: " + err.Error())
+	}
+
+	// Check FIT image existence and read it into memory
+	fitImage := mountPath + *fitFilePath
```
insomniacslk: this should use `path.Join` too

verifiedboot/main.go
```diff
@@ -0,0 +1,173 @@
...
+	fitImageData, err := ioutil.ReadFile(fitImage)
+	if err != nil {
+		recoverer.Recover("Can't read FIT image by given path: " + err.Error())
+	}
+
+	// Verify signature of FIT image on device
+	if *bootMode == booter.BootModeVerified || *bootMode == booter.BootModeBoth {
+		// Read fit image signature into memory
+		fitImageSignature := mountPath + *fitFilePath + SignatureFileExt
```
insomniacslk: `path.Join` here too

pkg/booter/verifiedbooter.go
```diff
@@ -0,0 +1,116 @@
...
+
+// VerifiedBooter implements the Booter interface for booting securely
+// into the operating system. This includes verified and measured boot
+// meachanisms.
+type VerifiedBooter struct {
+	Type       string `json:"type"`
+	BootMode   string `json:"boot_mode"`
+	DeviceUUID string `json:"device_uuid"`
+	FitFile    string `json:"fit_file"`
+	Debug      string `json:"debug"`
```
insomniacslk: this should be `bool`. The JSON parser will take care of setting this appropriately for you

insomniacslk (changes requested):
I haven't actually reviewed pkg/verifiedbooter.go, just verifiedboot/main.go . That's because I'd split this PR in two parts, one where we implement the verifiedbooter command, and one where we implement the verified booting logic on top. I need to think more on how to plug verified booting on top of local/net booting (wrapper? embedding?) so let's speed up this PR and let's just do the verifiedboot command.

pkg/booter/verifiedbooter.go
zaolin: Done

verifiedboot/main.go
zaolin: yeah

verifiedboot/main.go
zaolin: partly done, let's fix that finally afterwards

verifiedboot/main.go
zaolin: Done

verifiedboot/main.go
zaolin: okay let me think about it. Follow up

verifiedboot/main.go
zaolin: done

verifiedboot/main.go
zaolin: done

verifiedboot/main.go
```diff
@@ -78,91 +78,85 @@ func main() {
 	flag.Parse()
 	log.Print(banner)
 
-	var recoverer recovery.Recoverer
+	var RecoveryHandler recovery.Recoverer
```
insomniacslk: I actually meant to change the name in the `recovery` package

pkg/booter/verifiedbooter.go
```diff
@@ -0,0 +1,110 @@
...
+	log.Printf("VerifiedBooter: %+v", nb)
+	if nb.Type != "verifiedboot" {
+		return nil, fmt.Errorf("Wrong type for VerifiedBooter: %s", nb.Type)
+	}
+
+	if nb.BootMode != BootModeMeasured && nb.BootMode != BootModeVerified && nb.BootMode != BootModeBoth {
+		return nil, fmt.Errorf("False boot mode for VerifiedBooter: %s", nb.BootMode)
+	}
+
+	if _, err := uuid.FromString(nb.DeviceUUID); err != nil {
```
insomniacslk: what about using Fiano's `uuid` package? I wrote that code originally, and it's equivalent in terms of features that we need, plus we have control over. Find it at https://github.com/linuxboot/fiano/tree/master/uuid .

insomniacslk (commented):
Will review the rest on Monday

PR #29, opened 2018-06-11
Move features from localboot into generic diskutils package

zaolin:
* Move filesystem.go and blockdev.go
* Make it available for all booters

### DIFF
--- a/localboot/main.go
+++ b/localboot/main.go
@@ -6,7 +6,7 @@ import (
 	"path"
 	"syscall"
 
-	"github.com/insomniacslk/systemboot/pkg/diskutils"
+	"github.com/insomniacslk/systemboot/pkg/storage"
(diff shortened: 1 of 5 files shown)


insomniacslk:
This is good - thanks! There is only one thing that I'd like to be different, i.e. the package name. The intent of `localboot` has been intentionally away from the concept of "disk" (that's why "local" boot instead of "disk" boot). I would like to stick on this concept for the pkg library too.
Additionally I'm not a fan of "utils", "common" and similar naming, as they suggest "the place where we put things we don't know where to put".

Can you rename it to `pkg/storage`? If you have better ideas for the name feel free to suggest :)

zaolin:
Done

insomniacslk (approved):

PR #31, opened 2018-06-13
TPM1 extension and fixes

zaolin:
* Add ResetLock function
* Add Seal function
* Add Unseal function
* Fix parameters of ClearOwnership and TakeOwnership
* Set default constants
* Fix info structure parsing

### DIFF
--- a/pkg/tpm/tpm.go
+++ b/pkg/tpm/tpm.go
@@ -78,9 +78,9 @@ type Info struct {
 	TemporarilyDeactivated bool
 }
 
-func formatBoolFromSysfs(data []byte) string {
-	formatString := string(data)
-	return strings.TrimSuffix(formatString, "\n")
+func bytesToBool(data []byte) (bool, error) {
+	s := strings.TrimSuffix(string(data), "\n")
+	return strconv.ParseBool(s)
 }
 
 // NewTPM gets a new TPM handle struct with
@@ -162,19 +162,19 @@ func getInfo() (*Info, error) {
 		}
 	}
 
-	owned, err := strconv.ParseBool(formatBoolFromSysfs(ownedBytes))
+	owned, err := bytesToBool(ownedBytes)
 	if err != nil {
 		return nil, err
 	}
-	active, err := strconv.ParseBool(formatBoolFromSysfs(activeBytes))
+	active, err := bytesToBool(activeBytes)
 	if err != nil {
 		return nil, err
 	}
-	enabled, err := strconv.ParseBool(formatBoolFromSysfs(enabledBytes))
+	enabled, err := bytesToBool(enabledBytes)
 	if err != nil {
 		return nil, err
 	}
-	tempDeactivated, err := strconv.ParseBool(formatBoolFromSysfs(tempDeactivatedBytes))
+	tempDeactivated, err := bytesToBool(tempDeactivatedBytes)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/tpm/tpm.go
```diff
@@ -118,6 +134,16 @@ func getInfo() (*Info, error) {
 		return nil, err
 	}
 
+	manufacturePrefix := "Manufacturer: "
+	var manufactureID string
```
insomniacslk: typo: manufacture -> manufacturer

pkg/tpm/tpm.go
```diff
@@ -118,6 +134,16 @@ func getInfo() (*Info, error) {
...
+	for _, lineBytes := range bytes.Split(caps, []byte{'\n'}) {
+		line := string(lineBytes)
+		if strings.HasPrefix(line, manufacturePrefix) {
+			manufactureID = line[len(manufacturePrefix):]
+		}
+	}
+	manufacture := Manufactures[manufactureID]
```
insomniacslk: this will return an empty string if the manufacturer ID is unknown. Do we want to explicitly signal that this is unknown? E.g. by setting it to the string "< unknown >" ? (spaces added around "unknown" or github would eat them)

pkg/tpm/tpm.go
```diff
@@ -127,24 +153,25 @@ func getInfo() (*Info, error) {
 		}
 	}
 
-	owned, err := strconv.ParseBool(string(ownedBytes))
+	owned, err := strconv.ParseBool(string(ownedBytes[0]))
```
insomniacslk: was this failing because of the newline?

pkg/tpm/tpm1.go
```diff
@@ -157,10 +180,11 @@ func (t *TPM1) Close() {
 // Summary returns a string with formatted TPM information
 func (t TPM1) Summary() string {
```
insomniacslk, line 181: as discussed in an unrelated diff (in linuxboot/fiano), `Summary` is suboptimal and we'd better use JSON and (if necessary) introspection. I agree with that, and we can implement it in the next diff. Then we can remove Summary entirely

pkg/tpm/tpm.go
zaolin: Done

pkg/tpm/tpm.go
zaolin: Done

pkg/tpm/tpm.go
zaolin: yes

insomniacslk: ParseBool accepts 0/false/FALSE, 1/true/TRUE. I don't expect the sysfs interface to return funny things in the future, but just for more robustness I'd rather strip off the newline, make sure we have only one line, and convert it to int. It's a common and non-trivial operation so I'd refactor it out to a small function

pkg/tpm/tpm.go
```diff
@@ -46,22 +57,32 @@ type TPM interface {
...
 type Info struct {
+	Manufacturer           string
 	Specification          string
 	Owned                  bool
 	Active                 bool
 	Enabled                bool
 	TemporarilyDeactivated bool
 }
 
+func formatBoolFromSysfs(data []byte) string {
```
insomniacslk: We can simplify, make it like this:

```
func bytesToBool(data []byte) (bool, error) {
    s := strings.TrimSuffix(string(data), "\n")
    return strconv.ParseBool(s)
}
```

Then below just call it like
```
owned, err := bytesToBool(ownedBytes)
if err != nil {
    ...
}
```

insomniacslk (approved):

PR #34, opened 2018-06-18
Add function for getting mountpoint by device

zaolin:


pkg/storage/blockdev.go
```diff
@@ -190,3 +198,27 @@ func FilterEFISystemPartitions(devices []BlockDev) ([]BlockDev, error) {
 	}
 	return esps, nil
 }
+
+// GetMountpointByDevice gets the mountpoint by given
+// device name. Returns on first match
+func GetMountpointByDevice(devicePath string) (string, error) {
+	mountsPath, err := os.Open(LinuxMountsPath)
```
insomniacslk: rename `mountsPath` to something else (e.g. `file`, `fd` or similar) to reflect its type

pkg/storage/blockdev.go
```diff
@@ -190,3 +198,27 @@ func FilterEFISystemPartitions(devices []BlockDev) ([]BlockDev, error) {
...
+	if err != nil {
+		return "", err
+	}
+
+	defer mountsPath.Close()
+	scanner := bufio.NewScanner(mountsPath)
+	scanner.Split(bufio.ScanLines)
```
insomniacslk: you don't need this call because `bufio.ScanLines` is already the default splitter

pkg/storage/blockdev.go
```diff
@@ -190,3 +198,27 @@ func FilterEFISystemPartitions(devices []BlockDev) ([]BlockDev, error) {
...
+
+	var mountpoint string
+	for scanner.Scan() {
+		deviceInfo := strings.Split(scanner.Text(), LinuxMountsPathDelimiter)
```
insomniacslk: nit: use `strings.Fields` instead of `strings.Split`. It takes into account also consecutive spaces (not an issue with /proc/mounts but it's more robust in general)

pkg/storage/blockdev.go
```diff
@@ -190,3 +198,27 @@ func FilterEFISystemPartitions(devices []BlockDev) ([]BlockDev, error) {
...
+		if deviceInfo[0] == devicePath {
+			mountpoint = deviceInfo[1]
```
insomniacslk: just `return deviceInfo[1], nil` at this point (or `return &deviceInfo[1], nil` if we go for a string pointer)

pkg/storage/blockdev.go
```diff
@@ -190,3 +198,27 @@ func FilterEFISystemPartitions(devices []BlockDev) ([]BlockDev, error) {
...
+			break
+		}
+	}
+
+	return mountpoint, nil
```
insomniacslk: just `return "", nil` (or `return nil, nil` if we go for a string pointer)

pkg/storage/blockdev.go
```diff
@@ -190,3 +198,27 @@ func FilterEFISystemPartitions(devices []BlockDev) ([]BlockDev, error) {
...
+	mountsPath, err := os.Open(LinuxMountsPath)
+	if err != nil {
+		return "", err
+	}
+
+	defer mountsPath.Close()
+	scanner := bufio.NewScanner(mountsPath)
+	scanner.Split(bufio.ScanLines)
+
+	var mountpoint string
```
insomniacslk: you don't need this intermediate variable, see below

pkg/storage/blockdev.go
```diff
@@ -190,3 +198,27 @@ func FilterEFISystemPartitions(devices []BlockDev) ([]BlockDev, error) {
 	}
 	return esps, nil
 }
+
+// GetMountpointByDevice gets the mountpoint by given
+// device name. Returns on first match
+func GetMountpointByDevice(devicePath string) (string, error) {
```
insomniacslk: I am not convinced by this interface: if the device doesn't have a mount point, it will return the empty string, and no error. The caller could check for the empty string, but this is inconsistent.

Is the lack of a mount point to be considered an error? And if not, shall we rather return a `nil` string pointer and a `nil` error to say "no error, but no mountpoint either"?

pkg/storage/blockdev_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestFindMountPointNoError checks that there is no
+// error in parsong the test output.
+func TestFindMountPointNoError(t *testing.T) {
```
insomniacslk: this test is actually not necessary, as it is covered by the test below.

It would actually be useful to test the case where the device is not in the mounts file

insomniacslk (changes requested):

pkg/storage/blockdev.go
zaolin: Done

pkg/storage/blockdev.go
zaolin: Done

pkg/storage/blockdev.go
zaolin: Done

pkg/storage/blockdev.go
zaolin: Done

pkg/storage/blockdev.go
zaolin: Done

pkg/storage/blockdev.go
zaolin: Yeah let's use string pointer here

pkg/storage/blockdev.go
zaolin: Done

pkg/storage/blockdev_test.go
zaolin: Done

pkg/storage/blockdev_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestFindMountPointNotExists checks that non existent
+// entry is checked and nil returned
+func TestFindMountPointNotExists(t *testing.T) {
+	LinuxMountsPath = "tests/mounts"
+	mountpoint, _ := GetMountpointByDevice("/dev/mapper/sys-oldxxxxx")
+	require.Nil(t, mountpoint)
```
insomniacslk: this should become:
```
_, err := GetMountpointByDevice(...)
require.Error(t, err)
``

zaolin: Done

insomniacslk: this should be `require.Error`, see the example code I wrote in my previous comment

insomniacslk (approved):

PR #35, opened 2018-06-21
 Add TPM reseal support

zaolin:


pkg/tpm/tpm1.go
```diff
@@ -145,6 +145,21 @@ func (t *TPM1) SealData(locality byte, pcrs []int, data []byte, srkPassword stri
 	return sealed, nil
 }
 
+// ResealData seals data against a given pcrInfo map and srkPassword
```
insomniacslk, line 148: can you expand the comment saying what each input parameter is about?

pkg/tpm/tpm1.go
```diff
@@ -145,6 +145,21 @@ func (t *TPM1) SealData(locality byte, pcrs []int, data []byte, srkPassword stri
...
+func (t *TPM1) ResealData(locality byte, pcrInfo map[int][]byte, data []byte, srkPassword string) ([]byte, error) {
+	var srkAuth [20]byte
+	if srkPassword != "" {
+		srkAuth = sha1.Sum([]byte(srkPassword))
```
insomniacslk: this means that if srkPassword is empty, srkAuth will be empty. Is that what's intended?

insomniacslk (approved):

PR #36, opened 2018-06-26
Add crypto extensions [1/3]

zaolin:
* Add Data signing
* Add loading of private keys
* Generate keypair functionality
* ED25519 support


pkg/crypto/ed25519.go
```diff
@@ -0,0 +1,101 @@
...
+
+// LoadPublicKeyFromFile loads PEM formatted ED25519 public key from file.
+func LoadPublicKeyFromFile(publicKeyPath string) ([]byte, error) {
+	x509PEM, err := ioutil.ReadFile(publicKeyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse x509 PEM file
+	block, _ := pem.Decode(x509PEM)
```
insomniacslk: shouldn't you loop through the various blocks and try to decode  all of them until you find a PubKeyIdentifier?

pkg/crypto/ed25519.go
```diff
@@ -0,0 +1,101 @@
...
+// LoadPrivateKeyFromFile loads PEM formatted ED25519 private key from file.
+func LoadPrivateKeyFromFile(privateKeyPath string, password []byte) ([]byte, error) {
+	x509PEM, err := ioutil.ReadFile(privateKeyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse x509 PEM file
+	block, _ := pem.Decode(x509PEM)
+	if block == nil || block.Type != PrivKeyIdentifier {
```
insomniacslk: same as above

pkg/crypto/ed25519.go
```diff
@@ -0,0 +1,101 @@
...
+		Bytes: privKey,
+	}
+
+	var pubBlock = &pem.Block{
+		Type:  "PUBLIC KEY",
+		Bytes: pubKey,
+	}
+
+	var privateKey []byte
+	if password != nil || len(password) < 1 {
```
insomniacslk: you can simply check that `len(password) < 1`, it will also work when `password == nil` (as long as `password`'s type remains `[]byte`)

pkg/crypto/ed25519.go
zaolin: In theory yes but normally only one key should be inside a PEM encoded keyfile :)

pkg/crypto/ed25519.go
zaolin: maybe I can add a loop here

pkg/crypto/ed25519.go
zaolin: okay

zaolin:
Done

pkg/crypto/ed25519.go
```diff
@@ -0,0 +1,111 @@
...
+func LoadPublicKeyFromFile(publicKeyPath string) ([]byte, error) {
+	x509PEM, err := ioutil.ReadFile(publicKeyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse x509 PEM file
+	var block *pem.Block
+	for {
+		block, _ = pem.Decode(x509PEM)
```
insomniacslk: this will try to decode x509PEM again and again, so if there is no PubKeyIdentifier as the first block it will loop forever. It should use the second value returned by `pem.Decode`, something like:
```
var (
    block *pem.Block
    remaining []byte = x509PEM
)
for {
    block, remaining = pem.Decode(remaining)
    if block == nil {
        return nil, errors.New("Cannot decode PEM file")
    }
    if block.Type == PubKeyIdentifier {
        break
    }
}
```

pkg/crypto/ed25519.go
```diff
@@ -0,0 +1,111 @@
...
+func LoadPrivateKeyFromFile(privateKeyPath string, password []byte) ([]byte, error) {
+	x509PEM, err := ioutil.ReadFile(privateKeyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse x509 PEM file
+	var block *pem.Block
+	for {
+		block, _ = pem.Decode(x509PEM)
```
insomniacslk: same as above. This should be actually moved into a separate function and called by LoadPublicKeyFromFile and LoadPrivateKeyFromFile

pkg/crypto/ed25519.go
insomniacslk: So should we enforce that there is only one key in the PEM?

pkg/crypto/ed25519.go
insomniacslk: Tried on the playground, https://play.golang.org/p/ZUgdLfB5Fvw

insomniacslk (approved):

PR #43, opened 2018-08-06
Added BootConfig definition

insomniacslk:
This is the first step to build @zaolin's BootConfig at https://github.com/systemboot/systemboot/pull/37 .
This is the base for more flexible booters that are inspired to FIT but with a simpler syntax and a standard parser (BootConfig uses JSON, FIT uses a custom format that requires a custom parser).

### DIFF
--- a/pkg/bootconfig/bootconfig.go
+++ b/pkg/bootconfig/bootconfig.go
@@ -1,0 +1,21 @@
+package bootconfig
+
+import "encoding/json"
+
+// BootConfig is a general-purpose boot configuration. It draws some
+// characteristics from FIT but it's not compatible with it. It uses
+// JSON for interoperability.
+type BootConfig struct {
+	Name       string `json:"name"`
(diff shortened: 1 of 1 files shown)


pkg/bootconfig/bootconfig.go
```diff
@@ -0,0 +1,22 @@
...
+
+// BootConfig is a general-purpose boot configuration. It draws some
+// characteristics from FIT but it's not compatible with it. It uses
+// JSON for interoperability.
+type BootConfig struct {
+	Name       string `json:"name"`
+	Kernel     string `json:"kernel"`
+	Initramfs  string `json:"initramfs,omitempty"`
+	KernelArgs string `json:"kernel_args,omitempty"`
+	DeviceTree string `json:"devicetree"`
```
zaolin: missing omitempty

zaolin (approved):

PR #51, opened 2018-08-28
pkg/bootconfig: Added Manifest.GetBootConfig, and tests

insomniacslk:
This patch adds:
* a new Version field in the Manifest, so we can track breaking changes in the interface
* a custom UnmarshalJSON for BootConfig, that enforces the presence of mandatory fields
* tests

This is an additional step towards BootConfig support as designed in https://github.com/systemboot/systemboot/pull/37


pkg/bootconfig/bootconfig.go
```diff
@@ -13,6 +17,32 @@ type BootConfig struct {
 	DeviceTree string `json:"devicetree,omitempty"`
 }
 
+// UnmarshalJSON implements the json.Unmarshaler interface to enforce the
+// presence of required fields.
+func (bc *BootConfig) UnmarshalJSON(data []byte) error {
```
Natolumin: I'm not sure I agree with this: this is a package (bootconfig), and all fields in the struct are exported, so any API user would be able to create a struct without the required fields if they want, without going through the json decoding. To me this is a sign that these constraints on the struct could be encoded in a `Valid()` function, which could be then enforced at usage time by whatever else consumes the config.

Additionally, I'm not sure about the constraint on the name: while it is important for example for displaying a menu or allowing a user to chose a config in a list, not having the name doesn't really prevent booting if the bootconfig is passed directly to the booter

insomniacslk: That's a very good point - I wasn't too convinced by this either. While I wanted to avoid an explicit check, it's probably better than an implicit check that is only run in a few conditions.
I'll remove this function and add `IsValid` instead.

On the name - we need a way to identify the entry. I agree that the name is not a good idea - we could use an index instead? In the end the boot entries are nothing else than items in a list, and a zero-based index could be the most consistent way to go. Thoughts?

Natolumin: Having descriptive names if possible is always useful if there's a human in the loop, so I wouldn't remove the possibility of assigning names (and definitely display them in any UI if available), but not require them either

For selection however, the index is definitely a better way to go as far as I'm concerned; if we don't have say a curses interface or tab-completion, I don't want to have to type the full name in the cli to boot a non-default entry (again, thinking about UI here rather than API, since the API could just consume the full struct in the worst case)

insomniacslk: so be it - index for accessing, optional name for the human. Thanks!

insomniacslk: meanwhile - any opinion on the mandatory Version field in the manifest?

Natolumin: Sounded reasonable to me, always version your APIs :)

pkg/bootconfig/bootconfig.go
```diff
@@ -5,14 +5,24 @@ import "encoding/json"
 // BootConfig is a general-purpose boot configuration. It draws some
 // characteristics from FIT but it's not compatible with it. It uses
 // JSON for interoperability.
+// If you add or remove fields, remember to update UnmarshalJSON.
```
Natolumin: There's no unmarshaljson anymore

Natolumin (dismissed):

insomniacslk:
@Natolumin sorry to ping you again - needs to be approved again after every update

Natolumin (approved):

PR #52, opened 2018-08-30
Move to kexecbin

zaolin:


localboot/bootconfig.go
```diff
@@ -1,42 +1,27 @@
...
-	"github.com/u-root/u-root/pkg/kexec"
+	"github.com/u-root/u-root/pkg/kexecbin"
 )
 
 // BootConfig holds information to boot a kernel using kexec
 type BootConfig struct {
-	Kernel    *os.File
-	Initramfs *os.File
-	Cmdline   string
+	KernelFilePath string
```
insomniacslk, line 9: looks good! Can you just rename this to AbsoluteKernelPath and AbsoluteInitrdPath?

insomniacslk: as discussed in person I'll commit this change on your branch and merge

insomniacslk: Just realized that this will actually break current users of systemboot - we need a way to select kexec vs. kexecbin instead

PR #54, opened 2018-09-24
Added FromZip to unpack boot configurations from a zip file

insomniacslk:
Added support for signed and zipped boot configurations.

Context at https://github.com/systemboot/systemboot/pull/37


pkg/bootconfig/zipconfig.go
```diff
@@ -0,0 +1,130 @@
...
+		if len(signature) != ed25519.SignatureSize {
+			return nil, "", fmt.Errorf("Short read when reading signature: want %d bytes, got %d", ed25519.SignatureSize, len(signature))
+		}
+
+		// Verify the signature against the public key and the zip file bytes
+		if ok := ed25519.Verify(pubkey, zipbytes, signature); !ok {
+			return nil, "", fmt.Errorf("Invalid ed25519 signature for file %s", filename)
+		}
+	} else {
+		log.Printf("No public key specified, the ZIP file will be unpacked without verification")
```
Natolumin: If your zip is signed but you don't give a publickey for verification, this will consider the whole file including the signature if I understand correctly, which might lead to an invalid zip ?

pkg/bootconfig/zipconfig.go
```diff
@@ -0,0 +1,130 @@
...
+				return nil, "", err
+			}
+			if f.Name == "manifest.json" {
+				// parse the Manifest containing the boot configurations
+				manifest, err = NewManifest(buf)
+				if err != nil {
+					return nil, "", err
+				}
+			}
+			// FIXME how to get permissions stored in the zip file?
```
Natolumin: Each `File` object should contain a `FileHeader` with a `Mode` function. It might also be empty though as the info is not mandatory in a zip file afaik

pkg/bootconfig/zipconfig.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	tempDir, err := ioutil.TempDir(os.TempDir(), "bootconfig")
+	if err != nil {
+		return nil, "", err
+	}
+	log.Printf("Created temporary directory %s", tempDir)
+	var manifest *Manifest
+	for _, f := range r.File {
+		destination := path.Join(tempDir, f.Name)
+		// NOTE I don't know if it's possible to have a zero-length file name in
```
Natolumin: Well, at least some functions in the stdlib do check: https://golang.org/src/archive/zip/struct.go#L285

pkg/bootconfig/zipconfig.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// FromZip tries to extract a boot configuration from a ZIP file after verifying
+// its signature with the provided public key file. The signature is expected to
+// be appended to the ZIP file and have fixed length `ed25519.SignatureSize` .
+// The returned string argument is the temporary directory where the files were
+// extracted, if successful.
+// No decoder (e.g. JSON, ZIP) or other function parsing the input file is called
+// before verifying the signature.
+func FromZip(filename string, pubkeyfile *string) (*Manifest, string, error) {
```
Natolumin, line 45: This seems to do many things, do we have a usecase for extracting the whole zip without parsing the manifest, or parsing the manifest without extracting the whole zip ? If so we may want a manifest-extract-only function and an extract-everything function ?

pkg/bootconfig/zipconfig.go
```diff
@@ -0,0 +1,130 @@
...
+			if err != nil {
+				return nil, "", err
+			}
+			buf, err := ioutil.ReadAll(fd)
+			if err != nil {
+				return nil, "", err
+			}
+			if f.Name == "manifest.json" {
+				// parse the Manifest containing the boot configurations
+				manifest, err = NewManifest(buf)
```
Natolumin: So if we have multiple manifests in the zip, the latest one in the decompression order is the only one taken into account right ? Since it's not especially simple/well-known how to control file order in a zip file this might warrant a warning in documentation that you *probably* want to only have one manifest ever

Natolumin (dismissed):
I'm not sure if I understand everything and I'm definitely not qualified to validate the security flow

There are several things that warrant better documentation/comments though:

pkg/bootconfig/zipconfig.go
insomniacslk: You can append any arbitrary content to a ZIP file and it would still be extracted correctly, it's a known aspect of ZIP (and other file formats) that enables "features" like this. I can add a comment to explain it

pkg/bootconfig/zipconfig.go
insomniacslk: good point! I was looking at attributes, not methods, and this slipped through, thanks :)

pkg/bootconfig/zipconfig.go
insomniacslk: ok, I will check that too

pkg/bootconfig/zipconfig.go
insomniacslk: fair point. I don't think we have a use case to do unpacking without manifest extraction and optional signature verification. However I could still move the ZIP unpacking in its own `Unpack` method, and wrap it within `FromZip` (that will call `Unpack` and look for a valid manifest in the output directory). How does that sound?

pkg/bootconfig/zipconfig.go
insomniacslk: Yes - just checked that you can have duplicate files in a ZIP (and that, uh... "makes sense"), https://play.golang.org/p/kfvLvKxnzlP 

```
$ unzip -l duplicate.zip 
Archive:  duplicate.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        5  1980-00-00 00:00   duplicate
        6  1980-00-00 00:00   duplicate
---------                     -------
       11                     2 files
```

I will add a comment and a warning if a duplicate manifest is found.

pkg/bootconfig/zipconfig.go
insomniacslk: I don't think we should apply a default if the mode is empty though, as there is no way from a ZIP file to to tell if this is intentional or unspecified. I will just use `.Mode()` and let the user figure out

pkg/bootconfig/zipconfig.go
insomniacslk: on a second thought, I think I prefer this approach, and this can be split at a later stage if need be. I prefer to still be able to reuse the already-loaded-in-memory ZIP file when extracting it rather than discarding it and re-reading the file to extract it

Natolumin: Definitely makes sense to not reextract just to find the manifest if we've already extracted before. You'd need some heavier redesign to be able to just reuse the previously extracted image (at least store the location in tmpfs) and since we don't have a usecase right now let's add the complexity later

insomniacslk: thanks. If the other updates are good can you please re-approve?

Natolumin (approved):

PR #57, opened 2018-09-26
localboot: added grub and path mode, and registered localbooter for VPD

insomniacslk:
Improvements to local booting.
* Added -grub, -guid, -kernel, -initramfs, -cmdline options to select the boot method
* added LocalBooter and registered in the available booters
* updated uinit to use -grub as default (previously it was an implicit default)

### DIFF
--- a/localboot/main.go
+++ b/localboot/main.go
@@ -15,14 +15,14 @@ import (
 // TODO use a proper parser for grub config (see grub.go)
 
 var (
-	baseMountPoint = flag.String("m", "/mnt", "Base mount point where to mount partitions")
-	doDryRun       = flag.Bool("dryrun", false, "Do not actually kexec into the boot config")
-	doDebug        = flag.Bool("d", false, "Print debug output")
-	doGrub         = flag.Bool("grub", false, "Use GRUB mode, i.e. look for valid Grub/Grub2 configuration in default locations to boot a kernel. GRUB mode ignores -kernel/-initramfs/-cmdline")
-	kernelPath     = flag.String("kernel", "", "Specify the path of the kernel to execute. If using -grub, this argument is ignored")
-	initramfsPath  = flag.String("initramfs", "", "Specify the path of the initramfs to load. If using -grub, this argument is ignored")
-	kernelCmdline  = flag.String("cmdline", "", "Specify the kernel command line. If using -grub, this argument is ignored")
-	deviceGUID     = flag.String("guid", "", "GUID of the device where the kernel (and optionally initramfs) are located. Ignored if -grub is set or if -kernel is not specified")
+	flagBaseMountPoint = flag.String("m", "/mnt", "Base mount point where to mount partitions")
+	flagDryRun         = flag.Bool("dryrun", false, "Do not actually kexec into the boot config")
+	flagDebug          = flag.Bool("d", false, "Print debug output")
+	flagGrubMode       = flag.Bool("grub", false, "Use GRUB mode, i.e. look for valid Grub/Grub2 configuration in default locations to boot a kernel. GRUB mode ignores -kernel/-initramfs/-cmdline")
+	flagKernelPath     = flag.String("kernel", "", "Specify the path of the kernel to execute. If using -grub, this argument is ignored")
+	flagInitramfsPath  = flag.String("initramfs", "", "Specify the path of the initramfs to load. If using -grub, this argument is ignored")
+	flagKernelCmdline  = flag.String("cmdline", "", "Specify the kernel command line. If using -grub, this argument is ignored")
+	flagDeviceGUID     = flag.String("guid", "", "GUID of the device where the kernel (and optionally initramfs) are located. Ignored if -grub is set or if -kernel is not specified")
 )
 
 var debug = func(string, ...interface{}) {}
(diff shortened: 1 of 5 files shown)


localboot/main.go
```diff
@@ -2,56 +2,43 @@ package main
...
+// * look for every attached storage device
+// * try to mount every device using any of the kernel-supported filesystems
+// * look for a GRUB configuration in various well-known locations
+// * build a list of valid boot configurations from the found GRUB configuration files
+// * try to boot every valid boot configuration until one succeeds
+//
+// The first parameter, `mountedDevices` is a list of storage.Mountpoint
+// representing a mounted storage device.
+// The second parameter, `dryrun`, will not boot the found configurations if set
+// to true.
```
tfg13, line 73: nit: these seem to have changed names a bit :)

localboot/main.go
```diff
@@ -2,56 +2,43 @@ package main
...
+func BootGrubMode(devices []storage.BlockDev, baseMountpoint string, dryrun bool) error {
```
tfg13: baseMountpoint hides the outer var, not a huge issue, but maybe consider renaming it

localboot/main.go
```diff
@@ -96,10 +74,9 @@ func main() {
 
 	// try to kexec into every boot config kernel until one succeeds
 	for _, cfg := range bootconfigs {
-		log.Printf("trying to boot %s", cfg.Kernel.Name())
-		if *doDryRun {
-			// note: in dry-run, this loop brea at the first entry
-			// unconditionally
+		debug("Trying boot configuration %+v", cfg)
+		if dryrun {
```
tfg13: I find it confusing to go into the loop, and then break out immediately. I would put this if outside and never even go in in dryrun mode.

localboot/main.go
```diff
@@ -110,13 +87,116 @@ func main() {
...
+// BootPathMode tries to boot a kernel in PATH mode. This means:
+// * look for a partition with the given GUID and mount it
+// * look for the kernel and initramfs in the provided locations
+// * boot the kernel with the provided command line
+//
+// The first parameter, `devices`, is a list of `storage.BlockDev` objects to
+// look for the given GUID. The second parameter, `guid`, is the partition GUID
+// to look for. The third, `dryrun`, will not actually boot the found
+// configuration if set to true.
+func BootPathMode(devices []storage.BlockDev, baseMountpoint string, guid string, dryrun bool) error {
```
tfg13, line 159: See above, params have changed names, and baseMountpoint hides the outer var again

localboot/main.go
```diff
@@ -110,13 +87,116 @@ func main() {
...
+		Initramfs: initramfsFD,
+		Cmdline:   *kernelCmdline,
+	}
+	debug("Trying boot configuration %+v", cfg)
+	if dryrun {
+		log.Printf("Dry-run, will not actually boot")
+	} else {
+		if err := cfg.Boot(); err != nil {
+			cfg.Close()
+			return fmt.Errorf("Failed to boot kernel %s: %v", cfg.Kernel.Name(), err)
```
tfg13, line 193: Is there a chance we ever get here? Do you want to close the files from above before returning in this case? (kernelFD and initramfsFD)

pkg/booter/localbooter.go
```diff
@@ -0,0 +1,102 @@
...
+			"device_guid": "<device GUID or empty>"
+			"kernel": "<kernel path or empty>",
+			"kernel_args": "<kernel args or empty>",
+			"ramfs": "<ramfs path or empty>",
+		}
+
+		`type` is always set to "localboot"
+		`method` can be either "grub" or "path".
+		    The "grub" method will look for grub.cfg or grub2.cfg on the specified device.
+		    If no device is specified, it will look on all the attached storage devices,
```
tfg13, line 39: This sounds like you can use grub mode and still specify which device the grub is installed on. (but the code above always scans all devices instead) This feature would actually be very handy! For some big filesystems, it can even be very expensive to mount them all (double-digit _minutes_). In many cases, people using grub will know where its configuration is, so I think we should support this.

tfg13 (commented):
One feature request/comment, the rest is nits.

localboot/main.go
insomniacslk: thanks, fixed in the next commit :)

localboot/main.go
insomniacslk: good point, renamed all the flags to be `flag<NAME>`

localboot/main.go
insomniacslk: done

localboot/main.go
insomniacslk: yes, the boot config may fail and I want to close file descriptors that are not going to be used anymore

pkg/booter/localbooter.go
insomniacslk: That's an inconsistency, I will fix it, thanks for spotting!

insomniacslk: and will add a `guid` parameter to the GRUB mode too

insomniacslk:
Thanks @tfg13 . I've addressed all of your feedback except one thing in this first commit, and I will add the GUID flag in GRUB mode in the next commit

localboot/main.go
```diff
@@ -93,30 +119,127 @@ func main() {
...
+// * look for a partition with the given GUID and mount it
+// * look for the kernel and initramfs in the provided locations
+// * boot the kernel with the provided command line
+//
+// The first parameter, `devices` is a list of storage.BlockDev . The function
+// will look for bootable configurations on these devices
+// The second parameter, `baseMountPoint`, is the directory where the mount
+// points for each device will be created.
+// The third parameter, `guid`, is the partition GUID to look for.
+// The third parameter, `dryrun`, will not boot the found configurations if set
```
tfg13: *fourth

tfg13 (approved):

PR #61, opened 2018-10-25
Make the invalid fields message more chatty

rminnich:
Invalid number of fields

is not as useful as might seem :-)

Signed-off-by: Ronald G. Minnich <rminnich@gmail.com>


insomniacslk (approved):
Thanks @rminnich!

insomniacslk:
@rminnich see also https://github.com/systemboot/systemboot/pull/62 to fix the actual bug. There was a change in kernel 4.18

PR #62, opened 2018-10-25
Support blockdev stats for kernel 4.18+

insomniacslk:
Kernels 4.18+ have a different format for /sys/class/block/<device>/stat, four new fields appeared in https://github.com/torvalds/linux/commit/bdca3c87fb7ad1cc61d231d37eb0d8f90d001e0c . This patch addresses it, and relaxes the checks to look for at least 11 fields, not exactly 11 (or 15).

Thanks @rminnich for reporting it

### DIFF
--- a/pkg/storage/blockdev.go
+++ b/pkg/storage/blockdev.go
@@ -96,7 +96,7 @@ func BlockStatFromBytes(buf []byte) (*BlockStat, error) {
 	fields := strings.Fields(string(buf))
 	// BlockStat has 11 fields
 	if len(fields) < 11 {
-		return nil, fmt.Errorf("BlockStatFromBytes: parsing %q: got %d fields(%q), want 11", buf, len(fields), fields)
+		return nil, fmt.Errorf("BlockStatFromBytes: parsing %q: got %d fields(%q), want at least 11", buf, len(fields), fields)
 	}
 	intfields := make([]uint64, 0)
 	for _, field := range fields {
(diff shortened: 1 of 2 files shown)


insomniacslk:
cc @rminnich

rminnich (dismissed):
at some point if you want we can do this more simply with reflect.

insomniacslk:
> at some point if you want we can do this more simply with reflect.

I love and hate reflect, so I generally try to avoid it, but If you have a strong argument in favour I'll gladly use it

pkg/storage/blockdev.go
```diff
@@ -89,7 +95,7 @@ var SystemPartitionGUID = gpt.Guid([...]byte{
 func BlockStatFromBytes(buf []byte) (*BlockStat, error) {
 	fields := strings.Fields(string(buf))
 	// BlockStat has 11 fields
-	if len(fields) != 11 {
+	if len(fields) < 11 {
 		return nil, fmt.Errorf("BlockStatFromBytes: parsing %q: got %d fields(%q), want 11", buf, len(fields), fields)
```
tfg13:
```suggestion
		return nil, fmt.Errorf("BlockStatFromBytes: parsing %q: got %d fields(%q), want at least 11", buf, len(fields), fields)
```

tfg13 (dismissed):

tfg13 (approved):

PR #63, opened 2018-10-25
Require Go 1.11 in README

driusan:
The README for u-root says "Make sure your Go version is 1.11." The first step of the build process is to install github.com/u-root/u-root. Therefore, systemboot requires 1.11, not 1.10 or later

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ The `uinit` program just wraps `netboot` and `localboot` in a forever-loop logic
 
 ## How to build systemboot
 
-* Install a recent version of Go, we recommend 1.10 or later
+* Install Go 1.11
 * make sure that your PATH points appropriately to wherever Go stores the
   go-get'ed executables
 * Then build it with the `u-root` ramfs builder using the following commands:


insomniacslk:
Thanks for the contribution @driusan ! Before merging I want to check with @rjoleary what was the reason to use Go 1.11-only. The change was done in https://github.com/u-root/u-root/issues/926 over a month ago, but there's no explanation of the reason (which I assume there is).

cc @rminnich

rjoleary:
u-root switched to Go 1.11 just to keep up to date. We only run the CI tests with go1.11, which is why it's recommended in the readme. Everything may still work with go1.10, but there is no automated testing for it.

insomniacslk:
Thanks for the context @rjoleary ! At this point I would rather prefer to add systemboot builds for 1.10 and above rather than telling people they can only use 1.11. I am already making a PR for that

insomniacslk:
hey @driusan , now as per https://github.com/systemboot/systemboot/pull/64 we will support explicitly Go 1.10 and Go 1.11, despite u-root not running 1.10 in the CI. Can you please update your PR title and content to mention both versions? Thanks!

insomniacslk:
I guess we can close this PR. The current README is already consistent

PR #65, opened 2018-10-26
Unified pkg/bootconfig/BootConfig and localboot/BootConfig

insomniacslk:
There was duplication across these two packages, so I merged them into one BootConfig


localboot/grub.go
```diff
@@ -78,21 +79,11 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []BootConfig
...
 					cmdline = strings.Replace(cmdline, `\$`, "$", -1)
 				}
-				fullpath := path.Join(basedir, kernel)
-				fd, err := os.Open(fullpath)
-				if err != nil {
-					debug("error opening kernel file %s: %v", fullpath, err)
-				}
-				cfg.Kernel = fd
-				cfg.Cmdline = cmdline
+                cfg.Kernel = path.Join(basedir, kernel)
```
zaolin: indentation

localboot/main.go
```diff
@@ -134,8 +135,7 @@ func BootGrubMode(devices []storage.BlockDev, baseMountpoint string, guid string
 	for _, cfg := range bootconfigs {
 		debug("Trying boot configuration %+v", cfg)
 		if err := cfg.Boot(); err != nil {
-			log.Printf("Failed to boot kernel %s: %v", cfg.Kernel.Name(), err)
-			cfg.Close()
```
zaolin, line 138: ?

pkg/bootconfig/bootconfig.go
```diff
@@ -13,13 +20,49 @@ type BootConfig struct {
...
+        if kernel != nil {
+            if err := kernel.Close(); err != nil {
+                log.Printf("Error closing kernel file descriptor: %v", err)
+            }
+        }
+        if initramfs != nil {
+            if err := initramfs.Close(); err != nil {
+                log.Printf("Error closing initramfs file descriptor: %v", err)
+            }
+        }
```
zaolin: Don't forget the devicetree

localboot/grub.go
insomniacslk: darn, I committed from a machine without the goimports plugin -.-

localboot/main.go
insomniacslk: this is not necessary anymore, the fields are all strings now, not `os.File`, and the cleanup is done in `BootConfig.Boot`

pkg/bootconfig/bootconfig.go
insomniacslk: it's not necessasry here - the devicetree field is a string, and it's only used with KexecBin. The cleanup is done there

zaolin: ah right, kexec file load does not use the dt. Sorry!

localboot/main.go
zaolin: okay

localboot/grub.go
insomniacslk: also.. why doesn't Stickler complain?

zaolin (approved):

insomniacslk:
I've erroneously committed the kexecbin change along with this one, which was not tested yet and broke kexec if kexec-tools is not in the image :( will try to fix today, or roll-back the kexec change

insomniacslk:
The fix should be https://github.com/u-root/u-root/pull/996

insomniacslk:
second part of the fix at https://github.com/systemboot/systemboot/pull/67

PR #69, opened 2018-11-07
Remove TSS from systemboot

zaolin:
The TSS was moved to systemboot/tpmtool/pkg/tpm few months ago. Now we can remove the software stack from systemboot entirely.


insomniacslk:
Can you please update the description with the rationale? Otherwise it looks good to me

insomniacslk (approved):

Issue #46, opened 2018-08-07
Implement fallback brute-force booter


insomniacslk:
We want to offer an alternative booter that tries to boot from whatever available mean (dhcp, local disks, etc). This will be a command (e.g. `bruteforceboot` or something on those lines) that must be explicitly compiled in order to be used.

There are two scenarios:
* users who want a machine that always does the possible to boot
* users that want complete control on what a machine boots, and fail otherwise

Users of the first scenario will compile and use this booter, and users of the second scenario will simply not compile it, so it won't be used.

insomniacslk:
If compiled in, this booter will be called when all other entries configured in VPD fail, or no VPD entries exist. This is to address the case of a new machine with no configuration. Again, this is opt-in and only if the ramfs compiles this command in.

cc @Natolumin

PR #73, opened 2018-11-28
RFC: netboot from all interfaces

Natolumin:
For #46 we want to be able to boot from all interfaces if any of them is configurable and will offer us dhcp.
See the commit message for the more detailed rationale.

This is RFC because I'm not sure this is the best way of doing it - the alternative I think is having netboot take only one interface as it is now, but making uinit or whatever else calls netboot do the interface discovery. However that would prevent parallelizing bringing up and waiting for the interfaces, for which we have a 30s timeout right now

### DIFF
--- a/netboot/main.go
+++ b/netboot/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
@@ -16,12 +17,13 @@ import (
 	"github.com/insomniacslk/dhcp/iana"
 	"github.com/insomniacslk/dhcp/netboot"
 	"github.com/u-root/u-root/pkg/kexec"
+	"github.com/vishvananda/netlink"
 )
 
 var (
 	useV4              = flag.Bool("4", false, "Get a DHCPv4 lease")
 	useV6              = flag.Bool("6", true, "Get a DHCPv6 lease")
-	ifname             = flag.String("i", "eth0", "Interface to send packets through")
+	ifname             = flag.String("i", "", "Interface to send packets through")
 	dryRun             = flag.Bool("dryrun", false, "Do everything except assigning IP addresses, changing DNS, and kexec")
 	doDebug            = flag.Bool("d", false, "Print debug output")
 	skipDHCP           = flag.Bool("skip-dhcp", false, "Skip DHCP and rely on SLAAC for network configuration. This requires -netboot-url")
@@ -32,7 +34,7 @@ var (
 )
 
 const (
-	interfaceUpTimeout = 30 * time.Second
+	interfaceUpTimeout = 10 * time.Second
 )
 
 var banner = `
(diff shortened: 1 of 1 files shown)


netboot/main.go
```diff
@@ -62,134 +64,168 @@ func main() {
...
-					log.Fatalf("DHCPv6: cannot configure IPv6 addresses on interface %s: %v", *ifname, err)
-				}
-				// Set up DNS
-			}
-			if *overrideNetbootURL != "" {
-				bootfile = *overrideNetbootURL
+			log.Fatal("Could not obtain interfaces from netlink; aborting")
+		}
+		for _, l := range links {
+			if l.Attrs().Name == "lo" {
```
Natolumin: Note this is bad and I should check for the flag LOOPBACK instead, as it doesn't have to be named lo - will update this before final version

insomniacslk: yes, please. We can also add this as a separate function to the base library, as it would be used in other places too

insomniacslk: first attempt at https://github.com/insomniacslk/dhcp/pull/217, cc @pmazzini

Natolumin: I'm not sure this warrants a whole function, the abstraction provided by the netlink library is good enough for this to still be one line and explicit/readable. I just didn't think about it before sending the PR, it's not that it was more difficult than doing this

```go
l.Attrs().Flags & net.FlagLoopback
```

insomniacslk:
@Natolumin, I believe this is a sane default, and the new timeout is definitely better. Let's wait for the related PR before merging it

Natolumin:
Also I just realized seeing your related PR that I don't need netlink at all here to list the interfaces or even filter them, it's all possible with the standard library `net`
Still conflicted about the added value of the abstraction but now that it's there I'll use it

netboot/main.go
insomniacslk: the code is short and readable enough, I agree. However with a dedicated package I just wanted to make sure that users of these libraries get the loopback and non-loopback interfaces properly, and not, for example, by checking that the name is "lo"

insomniacslk:
> Also I just realized seeing your related PR that I don't need netlink at all here to list the interfaces or even filter them, it's all possible with the standard library net

yes, luckily Go's `net` is much better on that as it doesn't force us to use the Linux-specific netlink :)

> Still conflicted about the added value of the abstraction but now that it's there I'll use it

That PR is not merged yet

insomniacslk:
On the value of that PR - this would be used in the dhcpv4 and dhcpv6 package at least, and potentially here, with the goal of making sure that the interfaces are fetched correctly and not my name-matching. I don't have a strong opinion either though, there are pros and cons with both approaches

insomniacslk:
@Natolumin the `interfaces` PR is merged in the DHCP library, however feel free to avoid importing it here

insomniacslk (approved):

insomniacslk:
cc @zaolin - this PR will change the default behaviour of `netboot`. It will try DHCP sequentially on every interface, unless a specific NIC is configured via VPD

PR #75, opened 2018-11-30
uinit: Don't lose debug output in default cmds

Natolumin:
The quiet flag is used to decide between running the subcommands with
`-d` or not, but it doesn't make a difference since the output is
captured and discarded.

In the spirit of uinit being a "script", redirect the output of the
commands to stdout and stderr. This is done both when quiet and when
not, since the `-d` flag passed to the boot commands is already used to
increase or decrease verbosity


insomniacslk (approved):
looks good to me, thanks!

PR #70, opened 2018-11-07
Add TPM measurement support

zaolin:
* Add measurements via tpmtool pkg tpm
* Measure every data or code path
* Eventlog support is missing

PR #76, opened 2018-11-30
Add TPM measurement support

insomniacslk:
Replaces https://github.com/systemboot/systemboot/pull/70 , where I fucked up @zaolin 's commits :(

### DIFF
--- a/pkg/crypto/measure.go
+++ b/pkg/crypto/measure.go
@@ -21,39 +21,45 @@ const (
 // TryMeasureBootConfig measures bootconfig contents
 func TryMeasureBootConfig(name string, kernel string, initramfs string, kernelArgs string, deviceTree string) {
 	TPMInterface, err := tpm.NewTPM()
-	if err == nil {
-		TryMeasureData(BootConfig, []byte(name), &name)
-		TryMeasureData(BootConfig, []byte(kernel), &kernel)
-		TryMeasureData(BootConfig, []byte(initramfs), &initramfs)
-		TryMeasureData(BootConfig, []byte(kernelArgs), &kernelArgs)
-		TryMeasureData(BootConfig, []byte(deviceTree), &deviceTree)
-		TryMeasureFiles(kernel, initramfs, deviceTree)
-		TPMInterface.Close()
+	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/crypto/measure.go
```diff
@@ -0,0 +1,59 @@
...
+	// BootConfig type in PCR 8
+	BootConfig uint32 = 8
+	// ConfigData type in PCR 8
+	ConfigData uint32 = 8
+	// NvramVars type in PCR 9
+	NvramVars uint32 = 9
+)
+
+// TryMeasureBootConfig measures bootconfig contents
+func TryMeasureBootConfig(name string, kernel string, initramfs string, kernelArgs string, deviceTree string) {
```
Natolumin: nit: Wouldn't it be easier if this function was in bootconfig instead compared to having to unpack the struct to use it ? everything it uses is exported

Natolumin (dismissed):
Code looks reasonable, I'm not sure about the PCR numbers; I'd like @zaolin to have a look in any case

insomniacslk:
> Code looks reasonable, I'm not sure about the PCR numbers; I'd like @zaolin to have a look in any case

This is @zaolin 's original PR (#70), I just recreated it after I messed up a merge conflict

insomniacslk:
in other words - not my code. I will review it anyway once @zaolin has confirmed that this matches what he originally meant

zaolin:
I can confirm

pkg/crypto/measure.go
insomniacslk: I agree with you, this would probably better live in the bootconfig package

insomniacslk:
So there are three things:
1. @Natolumin 's advice on moving TryMeasureBootconfig to the `bootconfig` package
2. I'm wondering whether the measurement functions would be better suited to be part of `tpmtool`
3. could we reuse an open TPM rather than opening and closing it every time? This means making the measurements function as methods of a struct

CC @zaolin for further advice

insomniacslk:
@Natolumin , I think we can progress and merge this, and discuss on a separate issue. Could you please re-approve?

Natolumin (approved):

Issue #77, opened 2018-12-03
Fallback from kexecbin to u-root kexec failing


Natolumin:
When built without kexecbin and trying to localboot, I get this in logs:
```
Trying KexecBin on &{Name: Kernel:/mnt/sda1/vmlinuz Initramfs: KernelArgs:ro root=LABEL=/ biosdevname=0 net.ifnames=0 console=tty0 console=ttyS0,57600 DeviceTree:}
Failed to boot kernel /mnt/sda1/vmlinuz: exec: "kexecbin": executable file not found in $PATH
```

It seems that we get through [this](https://github.com/golang/go/blob/835f983da6866deec1c6550555c96de729c8848c/src/os/exec/lp_unix.go#L17) error path and error, instead of something that would satisfy the os.IsNotExist test in [bootconfig.go](https://github.com/systemboot/systemboot/blob/master/pkg/bootconfig/bootconfig.go#L36)

As a result we completely fail the boot instead of falling back to the kexec_file_load implementation from u-root

PR #79, opened 2018-12-03
Bootconfig fixes/improvements

Natolumin:
PR #65 refactored the bootconfig implementation, and in the process made it so we only load the kexec kernel, but don't boot it. This will be bundled with the fix for #77 since this code is unreachable because of that at the moment anyway

### DIFF
--- a/pkg/bootconfig/bootconfig.go
+++ b/pkg/bootconfig/bootconfig.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"log"
 	"os"
+	"os/exec"
 
 	"github.com/systemboot/systemboot/pkg/crypto"
 	"github.com/u-root/u-root/pkg/kexec"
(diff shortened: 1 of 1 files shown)


insomniacslk:
Thanks @Natolumin . WIP because testing, or because you plan to make more changes?

insomniacslk:
@Natolumin I've updated the branch via UI, you may need to `git pull` before making further changes

Natolumin:
both
I intend to add whatever changes we decide on for the kexecbin issue; I'll add a log of the error condition (this function should never return with `err==nil` since we're supposed to reboot if all goes well)

and separately I'm running into weird bugs with the booted kernel panicking that might be caused by us passing incorrect arguments

Natolumin:
Now I get the right error on boot, and it boots fine:
```
2018/12/04 18:16:33 BootConfig: KexecBin is not available, trying pure-Go kexec. Error: exec: "kexecbin": executable file not found in $PATH
```

insomniacslk:
From the title change I assume it's ready to review. @Natolumin do you confirm?

insomniacslk (approved):
Looks good to me, thanks!

Issue #87, opened 2018-12-20
Implement configtool program


insomniacslk:
See the one that was implemented in https://github.com/systemboot/systemboot/pull/37

PR #89, opened 2018-12-20
pkg/bootconfig: refactored, and added NewManifest

insomniacslk:
The old `NewManifest` function was a misnomer, it's a parser. Also added a proper `NewManifest` function, plus tests.


insomniacslk:
first step for #87

pmazzini (approved):
looks good to me

PR #91, opened 2019-01-11
Multiboot integration

zaolin:
Enables localboot to handle multiboot menu entries using kexeec support for multiboot introduced in [u-root PR 1021](https://github.com/u-root/u-root/pull/1021)

### DIFF
--- a/localboot/grub.go
+++ b/localboot/grub.go
@@ -3,43 +3,54 @@ package main
 import (
 	"io/ioutil"
 	"log"
+	"os"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/systemboot/systemboot/pkg/bootconfig"
 	"github.com/systemboot/systemboot/pkg/crypto"
 )
 
-// List of paths where to look for grub config files. Grub2Paths will look for
-// files with grub2-compatible syntax, GrubLegacyPaths similarly will treat
-// these as grub-legacy config files.
+// List of directories where to look for grub config files. The root dorectory
+// of each mountpoint, these folders inside the mountpoint and all subfolders
+// of these folders are searched
 var (
-	Grub2Paths = []string{
-		// grub2
-		"boot/grub2/grub.cfg",
-		"boot/grub2.cfg",
-		"grub2/grub.cfg",
-		"grub2.cfg",
-	}
-	GrubLegacyPaths = []string{
-		// grub legacy
-		"boot/grub/grub.cfg",
-		"boot/grub.cfg",
-		"grub/grub.cfg",
-		"grub.cfg",
+	GrubSearchDirectories = []string{
+		"boot",
+		"EFI",
+		"efi",
+		"grub",
+		"grub2",
 	}
 )
+
+type grubVersion int
+
+var (
+	grubV1 grubVersion = 1
+	grubV2 grubVersion = 2
+)
+
+func isGrubSearchDir(dirname string) bool {
+	for _, dir := range GrubSearchDirectories {
+		if dirname == dir {
+			return true
+		}
+	}
+	return false
+}
 
 // ParseGrubCfg parses the content of a grub.cfg and returns a list of
 // BootConfig structures, one for each menuentry, in the same order as they
 // appear in grub.cfg. All opened kernel and initrd files are relative to
 // basedir.
-func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.BootConfig {
+func ParseGrubCfg(ver grubVersion, grubcfg string, basedir string) []bootconfig.BootConfig {
 	// This parser sucks. It's not even a parser, it just looks for lines
 	// starting with menuentry, linux or initrd.
 	// TODO use a parser, e.g. https://github.com/alecthomas/participle
-	if grubVersion != 1 && grubVersion != 2 {
-		log.Printf("Warning: invalid GRUB version: %d", grubVersion)
+	if ver != grubV1 && ver != grubV2 {
+		log.Printf("Warning: invalid GRUB version: %d", ver)
 		return nil
 	}
 	bootconfigs := make([]bootconfig.BootConfig, 0)
@@ -65,6 +76,9 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
 			}
 			inMenuEntry = true
 			cfg = new(bootconfig.BootConfig)
+			name := strings.Join(sline[1:], " ")
+			name = strings.Split(name, "--")[0]
+			cfg.Name = name
 		} else if inMenuEntry {
 			// otherwise look for kernel and initramfs configuration
 			if len(sline) < 2 {
@@ -74,17 +88,32 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
 			if sline[0] == "linux" || sline[0] == "linux16" || sline[0] == "linuxefi" {
 				kernel := sline[1]
 				cmdline := strings.Join(sline[2:], " ")
-				if grubVersion == 2 {
-					// if grub2, unquote the string, as directives could be quoted
-					// https://www.gnu.org/software/grub/manual/grub/grub.html#Quoting
-					// TODO unquote everything, not just \$
-					cmdline = strings.Replace(cmdline, `\$`, "$", -1)
-				}
+				cmdline = unquote(ver, cmdline)
 				cfg.Kernel = path.Join(basedir, kernel)
 				cfg.KernelArgs = cmdline
 			} else if sline[0] == "initrd" || sline[0] == "initrd16" || sline[0] == "initrdefi" {
 				initrd := sline[1]
 				cfg.Initramfs = path.Join(basedir, initrd)
+			} else if sline[0] == "multiboot" || sline[0] == "multiboot2" {
+				multiboot := sline[1]
+				cmdline := strings.Join(sline[2:], " ")
+				cmdline = unquote(ver, cmdline)
+				cfg.Multiboot = path.Join(basedir, multiboot)
+				cfg.MultibootArgs = cmdline
+			} else if sline[0] == "module" || sline[0] == "module2" {
+				module := sline[1]
+				cmdline := strings.Join(sline[2:], " ")
+				if ver == grubV2 {
+					// if grub2, unquote the string, as directives could be quoted
+					// https://www.gnu.org/software/grub/manual/grub/grub.html#Quoting
+					// TODO unquote everything, not just \$
+					cmdline = strings.Replace(cmdline, `\$`, "$", -1)
+				}
+				module = path.Join(basedir, module)
+				if cmdline != "" {
+					module = module + " " + cmdline
+				}
+				cfg.Modules = append(cfg.Modules, module)
 			}
 		}
 	}
@@ -95,35 +124,53 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
 	return bootconfigs
 }
 
+func unquote(ver grubVersion, text string) string {
+	if ver == grubV2 {
+		// if grub2, unquote the string, as directives could be quoted
+		// https://www.gnu.org/software/grub/manual/grub/grub.html#Quoting
+		// TODO unquote everything, not just \$
+		return strings.Replace(text, `\$`, "$", -1)
+	}
+	// otherwise return the unmodified string
+	return text
+}
+
 // ScanGrubConfigs looks for grub2 and grub legacy config files in the known
 // locations and returns a list of boot configurations.
 func ScanGrubConfigs(basedir string) []bootconfig.BootConfig {
 	bootconfigs := make([]bootconfig.BootConfig, 0)
-	// Scan Grub 2 configurations
-	for _, grubpath := range Grub2Paths {
-		path := path.Join(basedir, grubpath)
-		log.Printf("Trying to read %s", path)
-		grubcfg, err := ioutil.ReadFile(path)
-		if err != nil {
-			log.Printf("cannot open %s: %v", path, err)
-			continue
+	err := filepath.Walk(basedir, func(currentPath string, info os.FileInfo, err error) error {
+		if path.Dir(currentPath) == basedir && info.IsDir() && !isGrubSearchDir(path.Base(currentPath)) {
+			log.Printf("Skip %s", currentPath)
+			return filepath.SkipDir // skip irrelevant toplevel directories
 		}
-		crypto.TryMeasureData(crypto.ConfigData, grubcfg, path)
-		cfgs := ParseGrubCfg(string(grubcfg), basedir, 2)
-		bootconfigs = append(bootconfigs, cfgs...)
-	}
-	// Scan Grub Legacy configurations
-	for _, grubpath := range GrubLegacyPaths {
-		path := path.Join(basedir, grubpath)
-		log.Printf("Trying to read %s", path)
-		grubcfg, err := ioutil.ReadFile(path)
-		if err != nil {
-			log.Printf("cannot open %s: %v", path, err)
-			continue
+		if info.IsDir() {
+			log.Printf("Check %s", currentPath)
+			return nil // continue
 		}
-		crypto.TryMeasureData(crypto.ConfigData, grubcfg, path)
-		cfgs := ParseGrubCfg(string(grubcfg), basedir, 1)
-		bootconfigs = append(bootconfigs, cfgs...)
+		cfgname := info.Name()
+		if cfgname == "grub.cfg" || cfgname == "grub2.cfg" {
+			var ver grubVersion
+			if cfgname == "grub.cfg" {
+				ver = grubV1
+			} else if cfgname == "grub2.cfg" {
+				ver = grubV2
+			}
+			// try parsing
+			log.Printf("Trying to read %s", currentPath)
+			grubcfg, errRead := ioutil.ReadFile(currentPath)
+			if errRead != nil {
+				log.Printf("cannot open %s: %v", currentPath, errRead)
+				return nil // continue anyway
+			}
+			crypto.TryMeasureData(crypto.ConfigDataPCR, grubcfg, currentPath)
+			cfgs := ParseGrubCfg(ver, string(grubcfg), basedir) // TODO get root dir for cfgs out of grub.cfg instead of taking the curren basedir
+			bootconfigs = append(bootconfigs, cfgs...)
+		}
+		return nil // continue
+	})
+	if err != nil {
+		log.Printf("filepath.Walk error: %v \n", err)
 	}
 	return bootconfigs
 }
--- a/localboot/grub.go
+++ b/localboot/grub.go
@@ -7,12 +7,15 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"unicode"
+
+	"golang.org/x/text/transform"
+	"golang.org/x/text/unicode/norm"
 
 	"github.com/systemboot/systemboot/pkg/bootconfig"
-	"github.com/systemboot/systemboot/pkg/crypto"
 )
 
-// List of directories where to look for grub config files. The root dorectory
+// List of directories where to recursively look for grub config files. The root dorectory
 // of each mountpoint, these folders inside the mountpoint and all subfolders
 // of these folders are searched
 var (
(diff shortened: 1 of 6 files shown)


insomniacslk:
@jensdrenhaus , @zaolin , can you please update the description?

insomniacslk:
also CC @lsiudut

insomniacslk:
Can you please add a short description?
Also, there is no testing of the bootconfig changes

localboot/grub.go
```diff
@@ -85,6 +87,31 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
 			} else if sline[0] == "initrd" || sline[0] == "initrd16" || sline[0] == "initrdefi" {
 				initrd := sline[1]
 				cfg.Initramfs = path.Join(basedir, initrd)
+			} else if sline[0] == "multiboot" {
+				multiboot := sline[1]
+				cmdline := strings.Join(sline[2:], " ")
+				if grubVersion == 2 {
```
insomniacslk: this block is duplicated from above. Can you please refactor into its own function?
Something like:

```
type grubVersion int

var (
	grubV1 grubVersion = 1
	grubV2 grubVersion = 2
)

func unquoteGrub(ver grubVersion, text string) {
	if ver == grubV2 {
		// if grub2, unquote the string, as directives could be quoted
		// https://www.gnu.org/software/grub/manual/grub/grub.html#Quoting
		// TODO unquote everything, not just \$
		return strings.Replace(text, `\$`, "$", -1)
	}
	return text
}
```

then use it like this:
```
cmdline = unquoteGrub(ver, cmdline)
```

Note that here I introduced a new type `grubVersion`, that has the same name a the variable passed as input of `ParseGrubCfg`. You should rename such variable into something different, e.g. just `ver`.

insomniacslk: I committed this change on your branch as I mentioned to @zaolin offline, feel free to revert it if you prefer to do it differently

pkg/crypto/measure.go
```diff
@@ -19,7 +19,7 @@ const (
 )
 
 // TryMeasureBootConfig measures bootconfig contents
-func TryMeasureBootConfig(name, kernel, initramfs, kernelArgs, deviceTree string) {
+func TryMeasureBootConfig(name, kernel, initramfs, kernelArgs, deviceTree, multiboot, multibootArgs string, modules []string) {
```
insomniacslk, line 22: This function should take a `BootConfig` reference as argument instead

insomniacslk:
Ping @jensdrenhaus - do you want me to commit on top of this branch?

pkg/crypto/measure.go
```diff
@@ -30,7 +30,12 @@ func TryMeasureBootConfig(name, kernel, initramfs, kernelArgs, deviceTree string
 	TryMeasureData(BootConfig, []byte(initramfs), initramfs)
```
insomniacslk: note: file names should not be measured, or at least should offer a sane default to skip name measurement. The risk with name measurements is that this becomes an operational pain with no real benefit, since the actual content is already verified. This can be done in another PR if you prefer

insomniacslk:
ping @jensdrenhaus

pkg/crypto/measure.go
jenstopp: I agree, but when import bootconfig to the crypto pkg there is a circle include. I think TryMeasureBootConfig() is redundant. When parsing the grub files and creating the bootconfigs TryMeasureData() ist used. So I reworked the part above to use TryMeasureData() as well.

Maybe it makes the crypto pkg more generic. However refactoring the crypto pkg schould be part of an other PR.

jenstopp:
@insomniacslk sorry for late response. I had to finish some testing on booting from efi systems based on the changes introduced here. Multiboot support here does work with MBR for now, but EFI support is about to come.

jenstopp:
ping @insomniacslk

insomniacslk:
can you rebase on top of master? In your last update you pulled a few unrelated commits

localboot/grub.go
```diff
@@ -3,43 +3,54 @@ package main
...
 
 	"github.com/systemboot/systemboot/pkg/bootconfig"
 	"github.com/systemboot/systemboot/pkg/crypto"
 )
 
-// List of paths where to look for grub config files. Grub2Paths will look for
-// files with grub2-compatible syntax, GrubLegacyPaths similarly will treat
-// these as grub-legacy config files.
+// List of directories where to look for grub config files. The root dorectory
+// of each mountpoint, these folders inside the mountpoint and all subfolders
```
insomniacslk, line 19: Just rephrase to "List of directories where to recursively look for grub config files"

localboot/grub.go
```diff
@@ -65,6 +76,9 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
 			}
 			inMenuEntry = true
 			cfg = new(bootconfig.BootConfig)
+			name := strings.Join(sline[1:], " ")
```
insomniacslk: this will panic on malformed lines, check that its length is at least 2 before this

localboot/grub.go
```diff
@@ -65,6 +76,9 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+			name = strings.Split(name, "--")[0]
```
insomniacslk: Is this to skip the optional arguments, I guess? https://www.gnu.org/software/grub/manual/grub/grub.html#menuentry 

This makes a lot of assumptions without taking into account, e.g., quoting. Parsing GRUB configs is a sh**show in general :(
For this specific part, we should improve `unquote` to also check single and double quotes (see https://www.gnu.org/software/grub/manual/grub/grub.html#Quoting  ) and implement a split method that handles quotes.

localboot/grub.go
```diff
@@ -74,17 +88,32 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+			} else if sline[0] == "multiboot" || sline[0] == "multiboot2" {
+				multiboot := sline[1]
+				cmdline := strings.Join(sline[2:], " ")
+				cmdline = unquote(ver, cmdline)
+				cfg.Multiboot = path.Join(basedir, multiboot)
+				cfg.MultibootArgs = cmdline
+			} else if sline[0] == "module" || sline[0] == "module2" {
+				module := sline[1]
+				cmdline := strings.Join(sline[2:], " ")
+				if ver == grubV2 {
```
insomniacslk: use `unquote` here

localboot/grub.go
```diff
@@ -95,35 +124,53 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
-	for _, grubpath := range Grub2Paths {
-		path := path.Join(basedir, grubpath)
-		log.Printf("Trying to read %s", path)
-		grubcfg, err := ioutil.ReadFile(path)
-		if err != nil {
-			log.Printf("cannot open %s: %v", path, err)
-			continue
+	err := filepath.Walk(basedir, func(currentPath string, info os.FileInfo, err error) error {
+		if path.Dir(currentPath) == basedir && info.IsDir() && !isGrubSearchDir(path.Base(currentPath)) {
+			log.Printf("Skip %s", currentPath)
```
insomniacslk: it's not necessary to log skipped directories

localboot/grub.go
```diff
@@ -95,35 +124,53 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
-	// Scan Grub Legacy configurations
-	for _, grubpath := range GrubLegacyPaths {
-		path := path.Join(basedir, grubpath)
-		log.Printf("Trying to read %s", path)
-		grubcfg, err := ioutil.ReadFile(path)
-		if err != nil {
-			log.Printf("cannot open %s: %v", path, err)
-			continue
+		if info.IsDir() {
+			log.Printf("Check %s", currentPath)
```
insomniacslk: this is also noise

localboot/grub.go
```diff
@@ -95,35 +124,53 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
-	// Scan Grub 2 configurations
-	for _, grubpath := range Grub2Paths {
-		path := path.Join(basedir, grubpath)
-		log.Printf("Trying to read %s", path)
-		grubcfg, err := ioutil.ReadFile(path)
-		if err != nil {
-			log.Printf("cannot open %s: %v", path, err)
-			continue
+	err := filepath.Walk(basedir, func(currentPath string, info os.FileInfo, err error) error {
+		if path.Dir(currentPath) == basedir && info.IsDir() && !isGrubSearchDir(path.Base(currentPath)) {
```
insomniacslk: `currentPath` should be normalized, and `err` should be checked as first thing

localboot/grub.go
```diff
@@ -95,35 +124,53 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
 	bootconfigs := make([]bootconfig.BootConfig, 0)
-	// Scan Grub 2 configurations
-	for _, grubpath := range Grub2Paths {
-		path := path.Join(basedir, grubpath)
-		log.Printf("Trying to read %s", path)
-		grubcfg, err := ioutil.ReadFile(path)
-		if err != nil {
-			log.Printf("cannot open %s: %v", path, err)
-			continue
+	err := filepath.Walk(basedir, func(currentPath string, info os.FileInfo, err error) error {
```
insomniacslk, line 153: `filepath.Walk` recurses the whole subtree, this is not a good idea because it may be very large. I would rather stick with a list of known paths

localboot/grub.go
```diff
@@ -95,35 +124,53 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+				ver = grubV2
+			}
+			// try parsing
+			log.Printf("Trying to read %s", currentPath)
+			grubcfg, errRead := ioutil.ReadFile(currentPath)
+			if errRead != nil {
+				log.Printf("cannot open %s: %v", currentPath, errRead)
+				return nil // continue anyway
+			}
+			crypto.TryMeasureData(crypto.ConfigDataPCR, grubcfg, currentPath)
```
insomniacslk: measuring boot configurations should not be part of parsing them

localboot/grub.go
```diff
@@ -95,35 +124,53 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+			cfgs := ParseGrubCfg(ver, string(grubcfg), basedir) // TODO get root dir for cfgs out of grub.cfg instead of taking the curren basedir
+			bootconfigs = append(bootconfigs, cfgs...)
 		}
-		crypto.TryMeasureData(crypto.ConfigData, grubcfg, path)
-		cfgs := ParseGrubCfg(string(grubcfg), basedir, 1)
-		bootconfigs = append(bootconfigs, cfgs...)
+		return nil // continue
```
insomniacslk:
```suggestion
		return nil
```

localboot/grub.go
```diff
@@ -95,35 +124,53 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+			}
+			// try parsing
+			log.Printf("Trying to read %s", currentPath)
+			grubcfg, errRead := ioutil.ReadFile(currentPath)
+			if errRead != nil {
+				log.Printf("cannot open %s: %v", currentPath, errRead)
+				return nil // continue anyway
+			}
+			crypto.TryMeasureData(crypto.ConfigDataPCR, grubcfg, currentPath)
+			cfgs := ParseGrubCfg(ver, string(grubcfg), basedir) // TODO get root dir for cfgs out of grub.cfg instead of taking the curren basedir
```
insomniacslk: comments before the line as per Go style guidelines

localboot/grub.go
```diff
@@ -95,35 +124,53 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+			bootconfigs = append(bootconfigs, cfgs...)
 		}
-		crypto.TryMeasureData(crypto.ConfigData, grubcfg, path)
-		cfgs := ParseGrubCfg(string(grubcfg), basedir, 1)
-		bootconfigs = append(bootconfigs, cfgs...)
+		return nil // continue
+	})
+	if err != nil {
+		log.Printf("filepath.Walk error: %v \n", err)
```
insomniacslk:
```suggestion
		log.Printf("filepath.Walk error: %v", err)
```

insomniacslk (commented):
There are multiple efforts to parse Grub here and in u-root, and I would like to unify them at some point (not in this PR of course), see 
https://github.com/u-root/u-root/tree/365523f459a9c603a90d89ab40ff1acc275024e8/pkg/diskboot 

CC @hugelgupf

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: Yor are right. In that way it works at least if singel quotes are present in the line.
Since the name field of the BootConfig struct wasn't set to a meaningfull value at all, I just want to assign something to be able to identify the choosen BootConfig in the log.
As mentioned above a well doing grub parser is an other topic. 
So I think we can go with it for now.

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: changed to `debug()`

localboot/grub.go
jenstopp: changed to debug()

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: done

pkg/crypto/measure.go
jenstopp: done

jenstopp (dismissed):

insomniacslk (changes requested):
you shouldn't approve your own commits :) please give me time to review. Discussed with @zaolin, I'm trying to fit this PR in between walls of meetings

localboot/grub.go
```diff
@@ -95,35 +126,64 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
-	for _, grubpath := range GrubLegacyPaths {
-		path := path.Join(basedir, grubpath)
-		log.Printf("Trying to read %s", path)
-		grubcfg, err := ioutil.ReadFile(path)
-		if err != nil {
-			log.Printf("cannot open %s: %v", path, err)
-			continue
+		t := transform.Chain(norm.NFD, transform.RemoveFunc(isMn), norm.NFC)
+		currentPath, _, _ = transform.String(t, currentPath)
+		if path.Dir(currentPath) == basedir && info.IsDir() && !isGrubSearchDir(path.Base(currentPath)) {
```
insomniacslk: move the `info.IsDir()` check as early as possible, it's an immediate indicator that we don't want a file, so we don't do additional operations unless necessary

localboot/grub.go
```diff
@@ -95,35 +126,64 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+			debug("Skip %s", currentPath)
+			// skip irrelevant toplevel directories
+			return filepath.SkipDir
 		}
-		crypto.TryMeasureData(crypto.ConfigData, grubcfg, path)
-		cfgs := ParseGrubCfg(string(grubcfg), basedir, 1)
-		bootconfigs = append(bootconfigs, cfgs...)
+		if info.IsDir() {
```
insomniacslk, line 159: remove this whole `if` block, you're already checking above if it's a directory

localboot/grub.go
```diff
@@ -95,35 +126,64 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+			debug("Check %s", currentPath)
+			// continue
+			return nil
+		}
+		cfgname := info.Name()
+		if cfgname == "grub.cfg" || cfgname == "grub2.cfg" {
```
insomniacslk: this block could be rewritten to be more clear. We also want an error if the grub file is found but cannot be read.
```
switch cfgname {
case "grub.cfg":
	ver = grubV1
case "grub2.cfg":
	ver = grubV2
default:
	return nil
}
log.Printf("Parsing %s", currentPath)
data, err := ioutil.ReadFile(currentPath)
if err != nil {
	return err
}
cfgs := ParseGrubCfg(ver, string(grubcfg), basedir)
bootconfigs = append(bootconfigs, cfgs...)
```

localboot/grub.go
```diff
@@ -95,35 +126,64 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+				ver = grubV1
+			} else if cfgname == "grub2.cfg" {
+				ver = grubV2
+			}
+			// try parsing
+			log.Printf("Trying to read %s", currentPath)
+			grubcfg, errRead := ioutil.ReadFile(currentPath)
+			if errRead != nil {
+				log.Printf("cannot open %s: %v", currentPath, errRead)
+				// continue anyway
```
insomniacslk: captured in the above comment - this should fail if we cannot read the file. What cases do you have in mind that would be OK to skip?

pkg/bootconfig/bootconfig.go
```diff
@@ -3,69 +3,102 @@ package bootconfig
...
 func (bc *BootConfig) IsValid() bool {
-	return bc.Kernel != ""
+	return (bc.Kernel != "" && bc.Multiboot == "") || (bc.Kernel == "" && bc.Multiboot != "")
 }
 
 // Boot tries to boot the kernel with optional initramfs and command line
 // options. If a device-tree is specified, that will be used too
 func (bc *BootConfig) Boot() error {
-	crypto.TryMeasureBootConfig(bc.Name, bc.Kernel, bc.Initramfs, bc.KernelArgs, bc.DeviceTree)
+	data := bc.Name + bc.Kernel + bc.Initramfs + bc.KernelArgs + bc.DeviceTree + bc.Multiboot + bc.MultibootArgs
```
insomniacslk, line 39: not measuring `bc.Modules`? We are not using `TryMeasureBootConfig` anymore?

localboot/grub.go
insomniacslk: discussed with @zaolin in person. We should find a middle ground here. Recursing this way is not a good idea, and Grub/Grub2 don't make this task easy - it's completely out of control where config files could be located. We can only make guesses unless we want implement a full-fledged parser (spoiler: no, we don't want).

However if we recursively walk a file tree, we need safeguards.

Also CC @marcoguerri

hugelgupf (commented):
Uh, why is there another grub parser?

insomniacslk:
> Uh, why is there another grub parser?

we have been using it for over a year, I'm pretty sure we talked a few times about it and about merging grub parsers into a single one

localboot/grub.go
jenstopp: Right. I've added a limitation to the directory depth.

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: done

localboot/grub.go
jenstopp: This looks way more elegant ;) - changed it

localboot/grub.go
jenstopp: I think that's a leftover from an early revision, where I just wanted to make sure every grub file gets found. But of cause, this should fail. I've fixed it.

pkg/bootconfig/bootconfig.go
jensdrenhaus: As you mentioned earlier

jenstopp: As you mentioned earlier regarding `TryMeasureBootconfig`:
> This function should take a `BootConfig` reference as argument instead

But therefor bootconfig pkg has to be imported in crypto pkg, which leads to circular import.

But what about adding a function like `func (bc* BootConfig) Strings() ([]string, error)` to the bootconfig pkg and change `TryMeasureBootConfig` to `func TryMeasureBootConfig(files ...string)` ?

tfg13:
This was merged as part of #106

Issue #94, opened 2019-01-18
Implement netboot retries


insomniacslk:
Support the retries flag in pkg/booters/netbooter.go (it's already implemented in the `netboot` command).

This will retry the whole DHCP + HTTP request. Also add a non-conditional retry (three times as a sane default?) for HTTP in case it fails for timeout or any other retryable error

insomniacslk:
Regarding DHCP retries, they are already implemented in https://github.com/systemboot/systemboot/blob/master/netboot/main.go#L34 and the default is 3. I will expose this number.

For the non-conditional HTTP retry I will go with the same default, and the retry will only happen in case of:
* timeout
* 500 Internal Server Error
* 502 Bad Gateway

I will not include 503 because this may hammer the backend servers without exponential backoff

insomniacslk:
cc @gepoggio

PR #99, opened 2019-02-05
netboot: added support for retryable HTTP failures

insomniacslk:
Fixes https://github.com/systemboot/systemboot/issues/94

CC @gepoggio

Relevant log from testing:
```
2019/02/05 12:02:02 Skipping DHCP
2019/02/05 12:02:02 DHCP: boot file URL is http://localhost/nbp
2019/02/05 12:02:02 DHCP: fetching boot file URL: http://localhost/nbp
2019/02/05 12:02:02 netboot: attempt 1 for http.Get
2019/02/05 12:02:02 retryable? net: false, http: true
2019/02/05 12:02:02 netboot: attempt 2 for http.Get
2019/02/05 12:02:02 retryable? net: false, http: false
2019/02/05 12:02:02 Cannot open TPM: open /dev/tpm0: permission denied
2019/02/05 12:02:02 DHCP: saved boot file to nbp
```

### DIFF
--- a/netboot/main.go
+++ b/netboot/main.go
@@ -37,6 +37,7 @@ var (
 
 const (
 	interfaceUpTimeout = 10 * time.Second
+	maxHTTPAttempts    = 3
 )
 
 var banner = `
@@ -109,6 +110,29 @@ func main() {
 	log.Fatalln("Could not boot from any interfaces")
 }
 
+func retryableNetError(err error) bool {
+	if err == nil {
+		return false
+	}
+	switch err := err.(type) {
+	case net.Error:
+		if err.Timeout() {
+			return true
+		}
+	}
+	return false
+}
+
+func retryableHTTPError(resp *http.Response) bool {
+	if resp == nil {
+		return false
+	}
+	if resp.StatusCode == 500 || resp.StatusCode == 502 {
+		return true
+	}
+	return false
+}
+
 func boot(ifname string, dhcp dhcpFunc) error {
 	var (
 		netconf  *netboot.NetConf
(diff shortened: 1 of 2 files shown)


netboot/main.go
```diff
@@ -37,6 +37,7 @@ var (
 
 const (
 	interfaceUpTimeout = 10 * time.Second
+	maxHTTPAttempts    = 3
```
pmazzini, line 40: Why not make it a variable similar to `dhcpRetries`?

pkg/booter/netbooter.go
```diff
@@ -66,6 +72,12 @@ func NewNetBooter(config []byte) (Booter, error) {
 // `netboot` command
 func (nb *NetBooter) Boot() error {
 	bootcmd := []string{"netboot", "-d", "-userclass", "linuxboot"}
+	if nb.OverrideURL != nil {
+		bootcmd = append(bootcmd, "-override-url", *nb.OverrideURL)
```
pmazzini: Should this be `netboot-url` instead to match the command in https://github.com/systemboot/systemboot/blob/master/netboot/main.go#L32?

pmazzini:
Commented. The rest LGTM.

netboot/main.go
insomniacslk: the original intent is to make it transparent to the user, rather than user-accessible. In my opinion this should be a standard feature, and I preferred not to expose it to the user here

pkg/booter/netbooter.go
insomniacslk: Good catch, yeah, it's a copy-pasta error, because override-url is used in the booter while netboot-url is in netboot

insomniacslk: updated in the last commit

netboot/main.go
pmazzini: Ok, I am good with it.

pmazzini (approved):

netboot/main.go
```diff
@@ -145,8 +169,16 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
 
 	log.Printf("DHCP: fetching boot file URL: %s", bootfile)
-	resp, err := http.Get(bootfile)
-	if err != nil {
+	var resp *http.Response
+	for attempt := 0; attempt < maxHTTPAttempts; attempt++ {
+		log.Printf("netboot: attempt %d for http.Get", attempt+1)
+		resp, err = http.Get(bootfile)
+		if err != nil && retryableNetError(err) || retryableHTTPError(resp) {
+			continue
```
insomniacslk, line 177: I just realized I didn't put a `Sleep` here. I'll make another PR

Issue #100, opened 2019-02-05
netboot: implement exponential backoff in retries


insomniacslk:
And once done, also cover other retriable errors, like HTTP 503

PR #101, opened 2019-02-06
netboot: added delay for HTTP retries

insomniacslk:
in https://github.com/systemboot/systemboot/pull/99#pullrequestreview-200535675 I realized this needs a delay between retries

### DIFF
--- a/netboot/main.go
+++ b/netboot/main.go
@@ -38,6 +38,7 @@ var (
 const (
 	interfaceUpTimeout = 10 * time.Second
 	maxHTTPAttempts    = 3
+	retryInterval      = time.Second
 )
 
 var banner = `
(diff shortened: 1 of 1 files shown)


netboot/main.go
```diff
@@ -174,6 +175,7 @@ func boot(ifname string, dhcp dhcpFunc) error {
 		log.Printf("netboot: attempt %d for http.Get", attempt+1)
 		resp, err = http.Get(bootfile)
 		if err != nil && retryableNetError(err) || retryableHTTPError(resp) {
+			time.Sleep(retryInterval)
```
pmazzini, line 178: Should the delay be increased after each retry?

insomniacslk: yes, it's part of https://github.com/systemboot/systemboot/issues/100 , and the reason why I didn't implement retry on 503 to avoid hammering the remote servers until exponential backoff is there. With timeouts and 500/502 I'm assuming it's safer to retry without backoff

pmazzini (approved):

PR #102, opened 2019-03-14
Add vpdbootmanager utility to manage boot entries

rctl:
This will add a simple CLI util to manage boot entries. 

**Add netboot entry ex**

```
~/go/src/github.com/systemboot/systemboot/util/vpd $ mkdir test/ro
~/go/src/github.com/systemboot/systemboot/util/vpd $ go run main.go boot --vpd-dir test add netboot dhcpv6 AA:BB:CC:DD:EE:FF
INFO[0000] Successfylly added boot entry.
~/go/src/github.com/systemboot/systemboot/util/vpd $ ls test/ro/
Boot0001
~/go/src/github.com/systemboot/systemboot/util/vpd $ cat test/ro/Boot0001
{"type":"netboot","method":"dhcpv6","mac":"AA:BB:CC:DD:EE:FF","retries":0}
```

**Add localboot entry ex**

```
~/go/src/github.com/systemboot/systemboot/util/vpd $ go run main.go boot --vpd-dir test add localboot path 1234 fake_kernel_file
INFO[0000] Successfylly added boot entry.
~/go/src/github.com/systemboot/systemboot/util/vpd $ ls test/ro/
Boot0001	Boot0002	Boot0003
~/go/src/github.com/systemboot/systemboot/util/vpd $ cat test/ro/Boot0003
{"type":"localboot","method":"path","device_guid":"1234","kernel":"fake_kernel_file"}
```

**Extra**

```
~/go/src/github.com/systemboot/systemboot/util/vpd $ go run main.go boot --vpd-dir test list
INFO[0000] Boot entries found:
INFO[0000] 	Boot0001
INFO[0000] 	Boot0002
INFO[0000] 	Boot0003
~/go/src/github.com/systemboot/systemboot/util/vpd $ go run main.go boot --vpd-dir test remove Boot0001
INFO[0000] Removed boot entry!
~/go/src/github.com/systemboot/systemboot/util/vpd $ go run main.go boot --vpd-dir test remove Boot0002
INFO[0000] Removed boot entry!
~/go/src/github.com/systemboot/systemboot/util/vpd $ go run main.go boot --vpd-dir test remove Boot0003
INFO[0000] Removed boot entry!
~/go/src/github.com/systemboot/systemboot/util/vpd $ ls test/ro/
~/go/src/github.com/systemboot/systemboot/util/vpd $
```

Note: --vpd-dir is used to override the default dir when testing, the default value when unspecified is that which is specified in `pkg/vpd`. 

int ref T41421367

### DIFF
--- a/cmds/vpdbootmanager/add.go
+++ b/cmds/vpdbootmanager/add.go
@@ -1,0 +1,131 @@
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"net"
+	"os"
+
+	"github.com/systemboot/systemboot/pkg/booter"
+	"github.com/systemboot/systemboot/pkg/vpd"
+)
+
+var dryRun = false
+
+func add(entrytype string, args []string) error {
+	var entry booter.Booter
+	var err error
+	switch entrytype {
+	case "netboot":
+		if len(args) < 2 {
+			return fmt.Errorf("You need to pass method and MAC address")
+		}
+		entry, err = parseNetbootFlags(args[0], args[1], args[2:])
+		if err != nil {
+			return err
+		}
+	case "localboot":
+		if len(args) < 1 {
+			return fmt.Errorf("You need to provide method")
+		}
+		entry, err = parseLocalbootFlags(args[0], args[1:])
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("Unknown entry type")
+	}
+	if dryRun {
+		b, err := json.Marshal(entry)
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(os.Stderr, "Using -dryrun, will not write any variable. Content of boot entry:")
+		fmt.Println(string(b))
+		return nil
+	}
+	return addBootEntry(entry)
+}
+
+func parseLocalbootFlags(method string, args []string) (*booter.LocalBooter, error) {
+	cfg := &booter.LocalBooter{
+		Type:   "localboot",
+		Method: method,
+	}
+	flg := flag.NewFlagSet("localboot", flag.ExitOnError)
+	flg.StringVar(&cfg.KernelArgs, "kernel-args", "", "additional kernel args")
+	flg.StringVar(&cfg.Initramfs, "ramfs", "", "path of ramfs to be used for kexec'ing into the target kernel.")
+	flg.StringVar(&vpd.VpdDir, "vpd-dir", vpd.VpdDir, "VPD dir to use")
+	flg.BoolVar(&dryRun, "dryrun", false, "only print values that would be set")
+
+	switch method {
+	case "grub":
+		flg.Parse(args)
+	case "path":
+		if len(args) < 2 {
+			return nil, fmt.Errorf("You need to pass DeviceGUID and Kernel path")
+		}
+		cfg.DeviceGUID = args[0]
+		cfg.Kernel = args[1]
+		flg.Parse(args[2:])
+	default:
+		return nil, fmt.Errorf("Method needs to be grub or path")
+	}
+	return cfg, nil
+}
+
+func parseNetbootFlags(method, mac string, args []string) (*booter.NetBooter, error) {
+	if method != "dhcpv4" && method != "dhcpv6" {
+		return nil, fmt.Errorf("Method needs to be either dhcpv4 or dhcpv6")
+	}
+
+	_, err := net.ParseMAC(mac)
+	if err != nil {
+		return nil, err
+	}
+
+	cfg := &booter.NetBooter{
+		Type:   "netboot",
+		Method: method,
+		MAC:    mac,
+	}
+
+	flg := flag.NewFlagSet("netboot", flag.ExitOnError)
+	overrideURL := flg.String("override-url", "", "an optional URL used to override the boot file URL used")
+	retries := flg.Int("retries", -1, "the number of times a DHCP request should be retried if failed.")
+	flg.BoolVar(&dryRun, "dryrun", false, "only print values that would be set")
+	flg.StringVar(&vpd.VpdDir, "vpd-dir", vpd.VpdDir, "VPD dir to use")
+	flg.Parse(args)
+
+	if *overrideURL != "" {
+		cfg.OverrideURL = overrideURL
+	}
+
+	if *retries != -1 {
+		cfg.Retries = retries
+	}
+
+	return cfg, nil
+}
+
+func addBootEntry(cfg booter.Booter) error {
+	data, err := json.Marshal(cfg)
+	if err != nil {
+		return err
+	}
+	for i := 1; i < vpd.MaxBootEntry; i++ {
+		key := fmt.Sprintf("Boot%04d", i)
+		if _, err := vpd.Get(key, false); err != nil {
+			if os.IsNotExist(err) {
+				if err := vpd.Set(key, data, false); err != nil {
+					return err
+				}
+				return nil
+			}
+			return err
+		}
+	}
+	return errors.New("Maximum number of boot entries already set")
+}
--- a/cmds/vpdbootmanager/main.go
+++ b/cmds/vpdbootmanager/main.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+const usage = `Usage:
+%s add [netboot [dhcpv6|dhcpv4] [MAC] | localboot [grub|path [Device GUID] [Kernel Path]]]
+
+Ex.
+add localboot grub
+add netboot dhcpv6 AA:BB:CC:DD:EE:FF
+
+Flags for netboot:
+
+-override-url - an optional URL used to override the boot file URL used
+-retries - the number of times a DHCP request should be retried if failed
+
+Flags for localboot:
+
+-kernel-args - additional kernel args
+-ramfs - path of ramfs to be used for kexec'ing into the target kernel
+
+Global flags:
+
+-vpd-dir - VPD dir to use
+
+`
+
+func main() {
+	if err := cli(os.Args[1:]); err != nil {
+		fmt.Printf(usage, os.Args[0])
+		fmt.Printf("Error: %s\n\n", err)
+		os.Exit(1)
+	}
+}
+
+func cli(args []string) error {
+	if len(args) < 1 {
+		return fmt.Errorf("You need to provide action")
+	}
+	switch args[0] {
+	case "add":
+		return add(args[1], args[2:])
+	}
+	return fmt.Errorf("Unrecognized action")
+}
--- a/cmds/vpdbootmanager/add_test.go
+++ b/cmds/vpdbootmanager/add_test.go
@@ -1,0 +1,170 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/systemboot/systemboot/pkg/booter"
+
+	"github.com/systemboot/systemboot/pkg/vpd"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseNetboot(t *testing.T) {
+	b, err := parseNetbootFlags("dhcpv4", "aa:bb:cc:dd:ee:ff", []string{})
+	require.NoError(t, err)
+	require.Equal(t, "netboot", b.Type)
+	require.Equal(t, "dhcpv4", b.Method)
+	require.Equal(t, "aa:bb:cc:dd:ee:ff", b.MAC)
+	require.Nil(t, b.OverrideURL)
+	require.Nil(t, b.Retries)
+}
+
+func TestParseNetbootWithFlags(t *testing.T) {
+	b, err := parseNetbootFlags("dhcpv4", "aa:bb:cc:dd:ee:ff", []string{
+		"-override-url",
+		"http://url",
+		"-retries",
+		"1",
+		"-vpd-dir",
+		"test",
+	})
+	require.NoError(t, err)
+	require.Equal(t, "http://url", *b.OverrideURL)
+	require.Equal(t, 1, *b.Retries)
+	require.Equal(t, "test", vpd.VpdDir)
+}
+
+func TestParseLocalboot(t *testing.T) {
+	b, err := parseLocalbootFlags("grub", []string{})
+	require.NoError(t, err)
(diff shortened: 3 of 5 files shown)


insomniacslk:
apologies for the delay, I will review this tomorrow PST

insomniacslk (changes requested):
hey @rctl, thanks for your PR! I should have provided you with more context before starting, so here's some. LinuxBoot has significant space constraints, and we are trying to minimize the binary size (Go 1.12 isn't helping in that sense). This means that, while we like and use logrus and cobra in other projects, they would increase the size of the final binary on the flash chip. Furthermore the systemboot tools are meant to be simple enough not to require the power of cobra, nor the flexibility of logrus (so far).
Another thing that I should have mentioned is that the vpd library in systemboot doesn't support writing to VPD variables, because the sysfs kernel interface doesn't support it (see https://github.com/systemboot/systemboot/blob/master/pkg/vpd/vpd.go#L40).

The things that I would recommend changing are:
- use `log` from the standard library instead of `logrus`
- use `flag` instead of `cobra` (yeah, I don't like `flag` either)
- rename the package from `vpd` to something else: this tool should be about boot management rather than its underlying implementation
- put libraries under `pkg` and commands under `cmds`. At the moment there is no `cmds`, but I plan to move `localboot` and `netboot` under that directory, so please go ahead and create it
- reimplement `vpd.Set` so that it can actually write to the VPD variable. We can discuss how this should be done


This PR also requires testing from a real LinuxBoot environment, you can use Qemu or we can use an internal test platform on real hw.

rctl:
I anticipated that that might be the case here 😉Context is always good, but now that I have more of it we can iterate on this. I'll go ahead and implement your suggestions here. In general I think no part of this tooling is actually lib-type, but rather pure cmd stuff. Lets discuss `vpd.Set`, I'll ping you on this 😄

rctl:
@insomniacslk alright, I have amended this now according to the new scope and removing the deps did cut the binsize in half down to 3.2MB from around 6MB :) I think that maybe we should do vpd.Set in different diff as not to bloat this one. So this one should be good to go in terms of the CLI part of this issue :)

zaolin (dismissed):
First of all, give the tool and packages some meaningful name. Otherwise, it's hard to understand for which purpose it is used.

rctl:
@zaolin what would be a more meaningful name, can you be more specific? 

The main package is because of golang mechanics (in reality we might call this just "boot" or "boot-entry"), and add package for subcommand "add" (we might want to extend with remove/list with other logics, that's why I want to decouple the adding part in its own subpackage).

insomniacslk:
Hey @rctl , I just re-emerged, catching up with this now.

> In general I think no part of this tooling is actually lib-type, but rather pure cmd stuff.

fair enough, we can split it later if needed.

> Lets discuss vpd.Set, I'll ping you on this

Sounds good

As per what @zaolin said on the name, I guess he could have added a bit of context. There is already a u-root command called `boot`, so the first issue is that those names will clash unless the ramfs is built with only one of them.
The second problem is that `boot` is not too descriptive of what the tool does. This tool creates boot entries in the VPD sections, much like `efibootmanager` does for UEFI's boot variables. I would recommend using the name `vpdbootmanager` so that it's obvious to most users what this tool is about

rctl:
Alright, I see, then changing the naming here makes much more sense. Will do that :)

rctl:
Amended the commit and updated the naming :)

cmds/vpdbootmanager/main.go
```diff
@@ -0,0 +1,30 @@
...
+Ex.
+boot add localboot grub
+
+boot add netboot dhcpv6 AA:BB:CC:DD:EE:FF
+
+Use -h to show flag options.
+`
+
+func main() {
+	if len(os.Args) < 2 {
```
insomniacslk: you can use `flag` instead of `os.Args` here and elsewhere. We already use it in other tools so it won't impact the file size in "busybox mode"

cmds/vpdbootmanager/main.go
```diff
@@ -0,0 +1,30 @@
...
+		fmt.Println(usage)
```
insomniacslk: if you use `flag` as per the above, you can override `flag.Usage`, see https://golang.org/pkg/flag/#pkg-variables

cmds/vpdbootmanager/addentry/netboot.go
```diff
@@ -0,0 +1,56 @@
+package addentry
```
insomniacslk: this is small enough that it could be part of the vpdbootmanager package instead of its own

cmds/vpdbootmanager/addentry/netboot.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"net"
+	"os"
+
+	"github.com/systemboot/systemboot/pkg/booter"
+	"github.com/systemboot/systemboot/pkg/vpd"
+)
+
+func addNetbootEntry() {
+	if len(os.Args) < 5 {
```
insomniacslk: I'd rather pass arguments to `addNetbootEntry`, so it can work in isolation instead of depending on the external environment

cmds/vpdbootmanager/addentry/netboot.go
```diff
@@ -0,0 +1,56 @@
...
+		fmt.Println(usage)
+		os.Exit(1)
```
insomniacslk: return an error here instead of exiting - the caller can decide whether this is fatal or not

cmds/vpdbootmanager/addentry/netboot.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	_, err := net.ParseMAC(os.Args[4])
+	if err != nil {
+		fmt.Printf("err: %s", err)
```
insomniacslk: same as above, just `return err` here and below

cmds/vpdbootmanager/addentry/add.go
```diff
@@ -0,0 +1,45 @@
...
+
+	}
+}
+
+func addBootEntry(data []byte) error {
+	for i := 1; i < 10; i++ {
+		key := fmt.Sprintf("Boot%04d", i)
+		if _, err := vpd.Get(key, true); err != nil {
+			if os.IsNotExist(err) {
+				if err := vpd.Set(key, data, true); err != nil {
```
insomniacslk: I think we want boot variables to be in the RW partition, so `false` instead of `true` here. cc @zaolin to double-check

cmds/vpdbootmanager/addentry/add.go
```diff
@@ -0,0 +1,45 @@
...
+	switch os.Args[2] {
+	case "netboot":
+		addNetbootEntry()
+	case "localboot":
+
+	}
+}
+
+func addBootEntry(data []byte) error {
+	for i := 1; i < 10; i++ {
```
insomniacslk: why 10? For how inefficient it may sound, we should iterate over the whole boot variables space, i.e. [0, 9999]. This value should go to a package-level variable, in `pkg/vpdbootmanager`

cmds/vpdbootmanager/addentry/add.go
```diff
@@ -0,0 +1,45 @@
...
+
+const usage = `Usage:
+boot add [netboot [dhcpv6|dhcpv4] [MAC] | localboot [grub|path [Device GUID] [Kernel Path]]]
+
+Use -h for flag options
+`
+
+// Cmd executes the command line parsing for the add command
+func Cmd() {
+	if len(os.Args) < 3 {
```
insomniacslk: I started reviewing from below (because `main` was there) - the comments below about flag vs os.Args apply here and in the other files

insomniacslk (commented):
A bunch of comments about what we discussed in person already. Can you also add tests?

cmds/vpdbootmanager/main.go
rctl: So stuff in `pkg` are using `os` and it's pretty well tied in (see `addBootEntry` for example). Until we decide to drop this in `pkg/vpd` I will leave it

rctl:
Alright, restructured a bit. Because we have conditional flag parsing (depending on which sub command we use we have different flagset and different length of arguments before that) it is hard to do more to get rid of args[] style passing things. I fixed the error handling and some other things, maybe there is more we can do here, but I think this should be short and sweet. After all this is just a supersimple cli. Feel free to have another pass :)

rctl:
Will add a test as well after we have put an ok on the current structure :)

cmds/vpdbootmanager/main.go
insomniacslk: It's not about not importing `os` (which is, in fact, imported by `flag` itself), it's about not handling the argument vector manually. You can capture the options with `os.Args` in main, but you should not use it anymore afterwards, and just pass the relevant arguments via function args

cmds/vpdbootmanager/main.go
```diff
@@ -0,0 +1,54 @@
...
+boot add netboot dhcpv6 AA:BB:CC:DD:EE:FF
+
+Use -h to show flag options.
+`
+
+func main() {
+	flag.Usage = func() {
+		fmt.Println(usage)
+	}
+	if err := parse(os.Args[1:]); err != nil {
```
insomniacslk: `parse` is not reflecting what the function does - it's actually parsing and executing the subcommand. This should be either renamed or split into parsing and execution

cmds/vpdbootmanager/main.go
```diff
@@ -0,0 +1,54 @@
...
+		flag.Usage()
+		fmt.Printf("Error: %s\n\n", err)
+		os.Exit(1)
```
insomniacslk, line 35: here and above we are mixing flag and os, we should not. The options are:
1) we use `flag` entirely, i.e. `flag.Parse()`, `flag.Args` and so on (with the additional benefit that you don't have to handle the parsing error yourself, it will detect it, print an error, and exit accordingly)
2) we don't use `flag` at all, we keep the parsing functions below, and just pass command arguments to the relevant functions.

If you go with 1), `flag` can handle positional arguments. Basically anything that comes after the first non-switch (anything that doesn't start with "-") argument, is captured and returned via `flag.Args()`, and you can pass `flag.Args()[1:]` to the subcommand handler.

If you go with 2), just keep your parsing functions and pass the arguments around, but no mixing with `flag`

insomniacslk (commented):
> Will add a test as well after we have put an ok on the current structure :)

I'd recommend adding tests as soon as possible, because they will force you to structure your code in a testable way, or it may be harder to change it later (in other words, adding them after agreeing the structure doesn't guarantee that you won't have to change it again because you find it's not easy to test)

cmds/vpdbootmanager/main.go
rctl: We cannot use `flag.Args` here (which I would have loved to use) because we have `FlagSet`s further down in the parsing line. As in, for `flag.Args` to work we need to do `flag.Parse()` which will stip the arguments needed to be kept for our `FlagSet`s. So I guess it's option 2 here, will amend with that.

rctl:
I've added some testcases now :) Coverage at 87.9%, think that is as far as we can take it

cmds/vpdbootmanager/main.go
```diff
@@ -0,0 +1,36 @@
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+const usage = `Usage:
+boot add [netboot [dhcpv6|dhcpv4] [MAC] | localboot [grub|path [Device GUID] [Kernel Path]]]
```
insomniacslk: use `os.Args[0]` instead of "boot" here

cmds/vpdbootmanager/add.go
```diff
@@ -0,0 +1,118 @@
...
+	return addBootEntry(entry)
+}
+
+func parseLocalbootFlags(method string, args []string) (*booter.LocalBooter, error) {
+	cfg := &booter.LocalBooter{
+		Type:   "localboot",
+		Method: method,
+	}
+	flg := flag.NewFlagSet("localboot", flag.ExitOnError)
+	flg.StringVar(&cfg.KernelArgs, "kernel-args", "", "additional kernel args")
```
insomniacslk, line 58: the sub-flags should be documented in the usage

cmds/vpdbootmanager/main.go
```diff
@@ -0,0 +1,36 @@
...
+
+Ex.
+boot add localboot grub
+
+boot add netboot dhcpv6 AA:BB:CC:DD:EE:FF
+
+Use -h to show flag options.
+`
```
insomniacslk, line 29: can you also add an option that just prints the JSON value without writing to VPD? It will help with non in-band changes

insomniacslk:
The build failure is due to some dependencies migrating to Go 1.12 while we were still on Go 1.11. It's fixed in https://github.com/systemboot/systemboot/pull/107 . Once merged it should be enough to rebase on master for the CI to succeed here

cmds/vpdbootmanager/add.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	default:
+		return fmt.Errorf("Unknown entry type")
+	}
+	if dryRun {
+		b, err := json.Marshal(entry)
+		if err != nil {
+			return err
+		}
+		fmt.Println("Using -dryrun, will not write any variable. Content of boot entry:")
```
insomniacslk: can you print this on stderr? So that the JSON output can be redirected if desired

cmds/vpdbootmanager/add.go
```diff
@@ -0,0 +1,131 @@
...
+			return fmt.Errorf("You need to provide method")
+		}
+		entry, err = parseLocalbootFlags(args[0], args[1:])
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("Unknown entry type")
+	}
+	if dryRun {
```
insomniacslk, line 40: this (global var) is one of the reasons why I'd have preferred to separate the generation of the entry from writing it, but I guess we can live with it

insomniacslk (commented):
just one last nit (stderr instead of stdout)  and this is OK to merge for me

cmds/vpdbootmanager/add.go
rctl: Generation and writing **is** separated. This global var is beacuse go `flag` package is not smart enough to handle "subcommand" case. So we need to use two flagsets (one for each subcommand (localboot/netboot), beacuse they have different flags). So, to handle this case and not use any smarter flag-parser that support subcommand we need to store "global" flags in variable.

insomniacslk (approved):

PR #103, opened 2019-03-14
Fix netboot for dhcpv4 having invalid argument type

rctl:
An invalid argument netboot is blocking all tests, this will fix that problem


pmazzini (approved):

pmazzini:
Thanks for fixing it!

PR #104, opened 2019-03-14
Add cacert support to netboot

BigValen:
This allows a user to stash a base64 encoded CA Cert into VPD variables, and use it to authenticate https connections for netbooting.

2019/03/14 06:40:23 DHCP: network configuration: &{Addresses:[{IPNet:{IP:2001:db8:1::1c Mask:ffffffffffffffff0000000000000000} PreferredLifetime:3000 ValidLifetime:4000}] DNSServers:[2001:4860:4860::8888 2001:4860:4860::8844] DNSSearchList:[] Routers:[]}
2019/03/14 06:40:23 DHCP: configuring network interface eth0 with &{[{{2001:db8:1::1c ffffffffffffffff0000000000000000} 3000 4000}] [2001:4860:4860::8888 2001:4860:4860::8844] [] []}
2019/03/14 06:40:23 DHCP: boot file for interface eth0 is https://[2001:db8:1::1]:443/nbp
2019/03/14 06:40:23 DHCP: boot file URL is https://[2001:db8:1::1]:443/nbp
2019/03/14 06:40:23 1948 bytes of VPD certs appended
2019/03/14 06:40:23 https client setup to use certs from VPD
2019/03/14 06:40:23 DHCP: fetching boot file URL: https://[2001:db8:1::1]:443/nbp
2019/03/14 06:40:23 netboot: attempt 1 for http.Get
2019/03/14 06:40:24 Cannot open TPM: stat /dev/tpm0: no such file or directory
2019/03/14 06:40:24 DHCP: saved boot file to nbp
2019/03/14 06:40:24 DHCP: kexec'ing into nbp

This can be tested by setting up a CAcert for the IP you are downloading from, and base64 encoding it:

```
CNF=/etc/pki/tls/openssl.cnf 
openssl genrsa  -out rootCA.key 4096
openssl req -x509 -new -nodes -key rootCA.key -sha256 -days 1024 -out rootCA.crt -subj "/C=US/ST=CA/O=MyOrg, Inc./CN=2001:db8:1::1"
openssl genrsa -out '2001:db8:1::1'.key 2048
openssl req -new -sha256 \
	-key 2001\:db8\:1\:\:1.key \
	-subj "/C=US/ST=CA/O=MyOrg, Inc./CN=2001:db8:1::1" \
	-out '2001:db8:1::1'.csr \
	-addext "subjectAltName = 'IP:2001:db8:1:0:0:0:0:1'" -extensions  SAN -config <(printf "[SAN]\nsubjectAltName='IP:2001:db8:1:0:0:0:0:1'\n"| cat $CNF -)
openssl x509 -req -in 2001\:db8\:1\:\:1.csr \
	-CA rootCA.crt -CAkey rootCA.key -CAcreateserial \
	-out 2001\:db8\:1\:\:1.crt -days 500 -sha256 \
	-extfile <(printf "subjectAltName='IP:2001:db8:1:0:0:0:0:1'\n")
base64  < rootCA.crt > rootCA.crt.base64
vpd -f build/coreboot.rom -i RW_VPD -O
vpd -f build/coreboot.rom -i RW_VPD -S ca_certs=rootCA.crt.base64
```
Things left to debate:
- some people might want to have certs that are RO from userspace, so they can't be changed. Others rotate them every few months. 
 - I have a VM test harness, and a simple TLS webserver that was useful for testing this. This should be uploaded too.

### DIFF
--- a/netboot/main.go
+++ b/netboot/main.go
@@ -1,6 +1,10 @@
 package main
 
 import (
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/base64"
+	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -33,6 +37,8 @@ var (
 	readTimeout        = flag.Int("timeout", 3, "Read timeout in seconds")
 	dhcpRetries        = flag.Int("retries", 3, "Number of times a DHCP request is retried")
 	userClass          = flag.String("userclass", "", "Override DHCP User Class option")
+	caCertFile         = flag.String("cacerts", "", "Base64 encoded CA cert file")
+	skipCertVerify     = flag.Bool("skip-cert-verify", false, "Don't authenticate https certs")
 )
 
 const (
@@ -150,7 +156,7 @@ func boot(ifname string, dhcp dhcpFunc) error {
 		}
 		debug("DHCP: network configuration: %+v", netconf)
 		if !*dryRun {
-			log.Printf("DHCP: configuring network interface %s", ifname)
+			log.Printf("DHCP: configuring network interface %s with %v", ifname, netconf)
 			if err = netboot.ConfigureInterface(ifname, netconf); err != nil {
 				return fmt.Errorf("DHCP: cannot configure interface %s: %v", ifname, err)
 			}
@@ -165,15 +171,28 @@ func boot(ifname string, dhcp dhcpFunc) error {
 	}
 	debug("DHCP: boot file URL is %s", bootfile)
 	// check for supported schemes
-	if !strings.HasPrefix(bootfile, "http://") {
-		return fmt.Errorf("DHCP: can only handle http scheme")
+	scheme, err := getScheme(bootfile)
+	if err != nil {
+		return fmt.Errorf("DHCP: cannot get scheme from URL: %v", err)
+	}
+	if scheme == "" {
+		return errors.New("DHCP: no valid scheme found in URL")
 	}
 
+	client, err := getClientForBootfile(bootfile)
+	if err != nil {
+		return fmt.Errorf("DHCP: cannot get client for %s: %v", bootfile, err)
+	}
 	log.Printf("DHCP: fetching boot file URL: %s", bootfile)
+
 	var resp *http.Response
 	for attempt := 0; attempt < maxHTTPAttempts; attempt++ {
 		log.Printf("netboot: attempt %d for http.Get", attempt+1)
-		resp, err = http.Get(bootfile)
+		req, err := http.NewRequest(http.MethodGet, bootfile, nil)
+		if err != nil {
+			return fmt.Errorf("could not build request for %s: %v", bootfile, err)
+		}
+		resp, err = client.Do(req)
 		if err != nil && retryableNetError(err) || retryableHTTPError(resp) {
 			time.Sleep(retryInterval)
 			continue
@@ -186,7 +205,7 @@ func boot(ifname string, dhcp dhcpFunc) error {
 	// FIXME this will not be called if something fails after this point
 	defer resp.Body.Close()
 	if resp.StatusCode != 200 {
-		return fmt.Errorf("Status code is not 200 OK: %d", resp.StatusCode)
+		return fmt.Errorf("status code is not 200 OK: %d", resp.StatusCode)
 	}
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
@@ -199,11 +218,11 @@ func boot(ifname string, dhcp dhcpFunc) error {
 	}
 	// extract file name component
 	if strings.HasSuffix(u.Path, "/") {
-		return fmt.Errorf("Invalid file path, cannot end with '/': %s", u.Path)
+		return fmt.Errorf("invalid file path, cannot end with '/': %s", u.Path)
 	}
 	filename := filepath.Base(u.Path)
 	if filename == "." || filename == "" {
-		return fmt.Errorf("Invalid empty file name extracted from file path %s", u.Path)
+		return fmt.Errorf("invalid empty file name extracted from file path %s", u.Path)
 	}
 	if err = ioutil.WriteFile(filename, body, 0400); err != nil {
 		return fmt.Errorf("DHCP: cannot write to file %s: %v", filename, err)
@@ -223,6 +242,84 @@ func boot(ifname string, dhcp dhcpFunc) error {
 		}
 	}
 	return nil
+}
+
+func getScheme(urlstring string) (string, error) {
+	u, err := url.Parse(urlstring)
+	if err != nil {
+		return "", err
+	}
+	scheme := strings.ToLower(u.Scheme)
+	if scheme != "http" && scheme != "https" {
+		return "", fmt.Errorf("URL scheme '%s' must be http or https", scheme)
+	}
+	return scheme, nil
+}
+
+func loadCaCerts() (*x509.CertPool, error) {
+	rootCAs, err := x509.SystemCertPool()
+	if err != nil {
+		return nil, err
+	}
+	if rootCAs == nil {
+		debug("certs: rootCAs == nil")
+		rootCAs = x509.NewCertPool()
+	}
+	base64Certs, err := ioutil.ReadFile(*caCertFile)
+	if err != nil {
+		return nil, fmt.Errorf("could not find cert file '%v' - %v", *caCertFile, err)
+	}
+	// TODO: Decide if this should also support compressed certs
+	// Might be better to have a generic compressed config API
+	decLen := base64.StdEncoding.DecodedLen(len(base64Certs))
+	certs := make([]byte, decLen)
+	n, err := base64.StdEncoding.Decode(certs, base64Certs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to append %s to RootCAs: %v", *caCertFile, err)
+	}
+	if ok := rootCAs.AppendCertsFromPEM(certs); !ok {
+		debug("No certs extracted from VPD, using system certs only")
+	} else {
+		debug("%d bytes of certs extrated from VPD", n)
+	}
+	return rootCAs, nil
+
+}
+
+func getClientForBootfile(bootfile string) (*http.Client, error) {
+	var client *http.Client
+	scheme, err := getScheme(bootfile)
+	if err != nil {
+		return nil, err
+	}
+
+	switch scheme {
+	case "https":
+		var config *tls.Config
+		if *skipCertVerify {
+			config = &tls.Config{
+				InsecureSkipVerify: true,
+			}
+		} else if *caCertFile != "" {
+			rootCAs, err := loadCaCerts()
+			if err != nil {
+				return nil, err
+			}
+			config = &tls.Config{
+				RootCAs: rootCAs,
+			}
+		}
+		tr := &http.Transport{TLSClientConfig: config}
+		client = &http.Client{Transport: tr}
+		debug("https client setup (use certs from VPD: %t, skipCertVerify %t)",
+			*skipCertVerify, *caCertFile != "")
+	case "http":
+		client = &http.Client{}
+		debug("http client setup")
+	default:
+		return nil, fmt.Errorf("Scheme %s is unsupported", scheme)
+	}
+	return client, nil
 }
 
 type dhcpFunc func(string) (*netboot.NetConf, string, error)


netboot/main.go
```diff
@@ -33,6 +37,8 @@ var (
 	readTimeout        = flag.Int("timeout", 3, "Read timeout in seconds")
 	dhcpRetries        = flag.Int("retries", 3, "Number of times a DHCP request is retried")
 	userClass          = flag.String("userclass", "", "Override DHCP User Class option")
+	caCertFile         = flag.String("cacerts", "/sys/firmware/vpd/rw/ca_certs", "Bas64 encoded CA cert file")
```
zaolin: I would use RO because otherwise someone else can modify RW with his custom CA list

netboot/main.go
```diff
@@ -139,6 +145,7 @@ func boot(ifname string, dhcp dhcpFunc) error {
 		netconf  *netboot.NetConf
 		bootfile string
 		err      error
+	    client  *http.Client
```
pallotron: have you ran gofmt on this? this doesn't seem aligned well.

netboot/main.go
```diff
@@ -165,15 +172,50 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
+		if ok := rootCAs.AppendCertsFromPEM(certs); !ok {
+			debug("No VPD certs appended, using system certs only")
+		} else {
+			debug("%d bytes of VPD certs appended", n)
+		}
+		config := &tls.Config{
+			InsecureSkipVerify: *skipCertVerify,
+			RootCAs:            rootCAs,
+		}
+		tr := &http.Transport{TLSClientConfig: config}
```
zaolin: Do we want to setup client-specific security requirements here?

netboot/main.go
```diff
@@ -165,15 +172,50 @@ func boot(ifname string, dhcp dhcpFunc) error {
 	}
 	debug("DHCP: boot file URL is %s", bootfile)
 	// check for supported schemes
-	if !strings.HasPrefix(bootfile, "http://") {
-		return fmt.Errorf("DHCP: can only handle http scheme")
+	matched, err := regexp.Match("^(http|https)://", []byte(bootfile))
```
pallotron: nit: can you make these two blocks a separate function which returns the httpclient? this boot function is getting white long already.

netboot/main.go
BigValen: I'd initially thought I wanted certs RW, because we do update them every few months. Seems we can set RO ones with flashrom, or via the BMC anyway, so I'll move them to RO.

netboot/main.go
BigValen: What do you mean ? Are there other client-specific requirements, or do you think they should be set elsewhere?

netboot/main.go
BigValen: Done now. Might have had a softtab problem..

netboot/main.go
BigValen: done!

netboot/main.go
```diff
@@ -33,6 +37,8 @@ var (
 	readTimeout        = flag.Int("timeout", 3, "Read timeout in seconds")
 	dhcpRetries        = flag.Int("retries", 3, "Number of times a DHCP request is retried")
 	userClass          = flag.String("userclass", "", "Override DHCP User Class option")
+	caCertFile         = flag.String("cacerts", "/sys/firmware/vpd/ro/ca_certs", "Base64 encoded CA cert file")
```
insomniacslk: I think we should not set a default here, because VPD is just one of the possible backends (NVRAM backend coming soon). Other LinuxBoot users are not using VPD at all. I'd rather force the user to specify the path, which for simplicity can be wrapped in uinit (the only component that assumes the presence of VPD)

netboot/main.go
```diff
@@ -165,15 +171,25 @@ func boot(ifname string, dhcp dhcpFunc) error {
 	}
 	debug("DHCP: boot file URL is %s", bootfile)
 	// check for supported schemes
-	if !strings.HasPrefix(bootfile, "http://") {
-		return fmt.Errorf("DHCP: can only handle http scheme")
+	matched, err := regexp.Match("^(http|https)://", []byte(bootfile))
```
insomniacslk: we should probably consider mixed and upper case too

netboot/main.go
```diff
@@ -165,15 +171,25 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
+	if !matched || err != nil {
+		return fmt.Errorf("DHCP: can only handle http/https scheme")
```
insomniacslk: It's worth printing the error, if any. What about a more explicit:
```
if err != nil {
	return fmt.Errorf("DHCP: regexp matching failed: %v", err)
}
if !matched {
	return errors.New("DHCP: can only handle http/https schemes")
}
```

netboot/main.go
```diff
@@ -165,15 +171,25 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
 	}
 
+	client, err := getClientForBootfile(bootfile)
+	if err != nil {
+		return fmt.Errorf("DHCP: cannot get http client for %s", bootfile)
```
insomniacslk:
```suggestion
		return fmt.Errorf("DHCP: cannot get http client for %s: %v", bootfile, err)
```

netboot/main.go
```diff
@@ -165,15 +171,25 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
+	}
 	log.Printf("DHCP: fetching boot file URL: %s", bootfile)
+
 	var resp *http.Response
 	for attempt := 0; attempt < maxHTTPAttempts; attempt++ {
 		log.Printf("netboot: attempt %d for http.Get", attempt+1)
-		resp, err = http.Get(bootfile)
+		req, err := http.NewRequest(http.MethodGet, bootfile, nil)
+		if err != nil {
+			return fmt.Errorf("Could not build %s request for %s", http.MethodGet, bootfile)
```
insomniacslk:
```suggestion
			return fmt.Errorf("Could not build %s request for %s: %v", http.MethodGet, bootfile, err)
```

netboot/main.go
insomniacslk: I see that the same matching is done below in `getClientForBootfile`, so this should be refactored in a function. What about something like this? https://play.golang.org/p/2WAX_oHVg0m .

```

func getScheme(urlstring string) (string, error) {
	u, err := url.Parse(urlstring)
	if err != nil {
		return "", err
	}
	scheme := strings.ToLower(u.Scheme)
	if scheme != "http" && scheme != "https" {
		return "", errors.New("Invalid or unsupported URL scheme. Currently only http and https are supported")
	}
	return scheme, nil
}
```

netboot/main.go
```diff
@@ -225,6 +241,40 @@ func boot(ifname string, dhcp dhcpFunc) error {
 	return nil
 }
 
+func getClientForBootfile(bootfile string) (*http.Client, error) {
+	var client *http.Client
+	if strings.HasPrefix(bootfile, "https://") {
```
insomniacslk: this should also use `getScheme` as suggested above`

netboot/main.go
```diff
@@ -225,6 +241,40 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
+			debug("%d bytes of VPD certs appended", n)
+		}
+		config := &tls.Config{
+			InsecureSkipVerify: *skipCertVerify,
+			RootCAs:            rootCAs,
+		}
+		tr := &http.Transport{TLSClientConfig: config}
+		client = &http.Client{Transport: tr}
+		debug("https client setup to use certs from VPD")
+	} else {
```
insomniacslk: the `else` should be defensively explicit to avoid errors when we will support more schemes

netboot/main.go
BigValen: Of course, sounds good.

netboot/main.go
BigValen: Good call.

netboot/main.go
BigValen: getScheme now returns an error if we don't try http or https, so it'll return before this if. If someone has added a new scheme to getScheme that getClientForBootfile doesn't, it's now a fatal error.

netboot/main.go
```diff
@@ -150,7 +156,7 @@ func boot(ifname string, dhcp dhcpFunc) error {
 		}
 		debug("DHCP: network configuration: %+v", netconf)
 		if !*dryRun {
-			log.Printf("DHCP: configuring network interface %s", ifname)
+			log.Printf("DHCP: configuring network interface %s with %v", ifname, netconf)
```
insomniacslk, line 159: nit: `%+v` also prints the names of the fields, it may help debugging

```suggestion
			log.Printf("DHCP: configuring network interface %s with %+v", ifname, netconf)
```

netboot/main.go
```diff
@@ -165,15 +171,25 @@ func boot(ifname string, dhcp dhcpFunc) error {
 	}
 	debug("DHCP: boot file URL is %s", bootfile)
 	// check for supported schemes
-	if !strings.HasPrefix(bootfile, "http://") {
-		return fmt.Errorf("DHCP: can only handle http scheme")
+	scheme, err := getScheme(bootfile)
+	if scheme == "" || err != nil {
```
insomniacslk: I would check the error and the value of scheme separately:
```
if err != nil {
	return fmt.Errorf("DHCP: cannot get scheme from URL: %v", err)
}
if scheme == "" {
	return errors.New("DHCP: no valid scheme found in URL")
}
```

netboot/main.go
```diff
@@ -165,15 +171,25 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
+		return errors.New("DHCP: can only handle http/https scheme")
 	}
 
+	client, err := getClientForBootfile(bootfile)
+	if err != nil {
+		return fmt.Errorf("DHCP: cannot get http client for %s: %v", bootfile, err)
```
insomniacslk: we may extend to non-HTTP in the future

```suggestion
		return fmt.Errorf("DHCP: cannot get client for %s: %v", bootfile, err)
```

netboot/main.go
```diff
@@ -165,15 +171,25 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
+	}
 	log.Printf("DHCP: fetching boot file URL: %s", bootfile)
+
 	var resp *http.Response
 	for attempt := 0; attempt < maxHTTPAttempts; attempt++ {
 		log.Printf("netboot: attempt %d for http.Get", attempt+1)
-		resp, err = http.Get(bootfile)
+		req, err := http.NewRequest(http.MethodGet, bootfile, nil)
```
insomniacslk, line 191: you reuse `http.MethodGet` twice, this would be better off in its own variable `method := http.MethodGet` used here and below

netboot/main.go
```diff
@@ -225,6 +241,82 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
+}
+
+func getClientForBootfile(bootfile string) (*http.Client, error) {
+	var client *http.Client
+	scheme, err := getScheme(bootfile)
+	if err != nil {
+		return nil, err
+	}
+
+	if scheme == "https" {
```
insomniacslk: nit: this could be a `switch`

netboot/main.go
```diff
@@ -225,6 +241,82 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
+			if err != nil {
+				return nil, err
+			}
+			config = &tls.Config{
+				RootCAs: rootCAs,
+			}
+		}
+		tr := &http.Transport{TLSClientConfig: config}
+		client = &http.Client{Transport: tr}
+		debug("https client setup to use certs from VPD")
```
insomniacslk: should this debug line also say whether we are skipping certificate verification?

insomniacslk (commented):
minor things, it looks good to me otherwise

netboot/main.go
BigValen: We actually print netconf on the line above, if -d is set - no need to add more to the logs here.

netboot/main.go
BigValen: I'll actually make the log message less verbose. We shouldn't care that it's a GET method.

netboot/main.go
```diff
@@ -225,6 +244,84 @@ func boot(ifname string, dhcp dhcpFunc) error {
...
+		}
+		tr := &http.Transport{TLSClientConfig: config}
+		client = &http.Client{Transport: tr}
+		debug("https client setup (use certs from VPD: %t, skipCertVerify %t)",
+			*skipCertVerify, *caCertFile != "")
+	case "http":
+		client = &http.Client{}
+		debug("http client setup")
+	default:
+		log.Fatalf("Scheme %s is unsupported", scheme)
```
insomniacslk: this should return an error instead of exiting with Fatal

insomniacslk (commented):
looks good! Just one last thing before merging

insomniacslk (approved):
lgtm! Thanks @BigValen !

PR #105, opened 2019-03-20
localboot: add config flag to select bootconfig

jenstopp:
A little feature that comes in handy mainly for debugging.
You can do ```localboot -grub -dryrun``` which prints you the found bootconfigs at the end:
pic
Then you can run ```localboot -config 1``` to use the corresponding bootconfig.

### DIFF
--- a/localboot/main.go
+++ b/localboot/main.go
@@ -19,6 +19,7 @@ var (
 	flagBaseMountPoint = flag.String("m", "/mnt", "Base mount point where to mount partitions")
 	flagDryRun         = flag.Bool("dryrun", false, "Do not actually kexec into the boot config")
 	flagDebug          = flag.Bool("d", false, "Print debug output")
+	flagConfigIdx      = flag.Int("config", -1, "Specify the index of the configuration to boot. The order is determined by the menu entries in the Grub config")
 	flagGrubMode       = flag.Bool("grub", false, "Use GRUB mode, i.e. look for valid Grub/Grub2 configuration in default locations to boot a kernel. GRUB mode ignores -kernel/-initramfs/-cmdline")
 	flagKernelPath     = flag.String("kernel", "", "Specify the path of the kernel to execute. If using -grub, this argument is ignored")
 	flagInitramfsPath  = flag.String("initramfs", "", "Specify the path of the initramfs to load. If using -grub, this argument is ignored")
@@ -72,7 +73,7 @@ func mountByGUID(devices []storage.BlockDev, filesystems []string, guid, baseMou
 // instead.
 // The fourth parameter, `dryrun`, will not boot the found configurations if set
 // to true.
-func BootGrubMode(devices []storage.BlockDev, baseMountpoint string, guid string, dryrun bool) error {
+func BootGrubMode(devices []storage.BlockDev, baseMountpoint string, guid string, dryrun bool, configIdx int) error {
 	// get a list of supported file systems for real devices (i.e. skip nodev)
 	debug("Getting list of supported filesystems")
 	filesystems, err := storage.GetSupportedFilesystems()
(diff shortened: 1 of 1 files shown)


zaolin (dismissed):

localboot/main.go
```diff
@@ -19,6 +19,7 @@ var (
 	flagBaseMountPoint = flag.String("m", "/mnt", "Base mount point where to mount partitions")
 	flagDryRun         = flag.Bool("dryrun", false, "Do not actually kexec into the boot config")
 	flagDebug          = flag.Bool("d", false, "Print debug output")
+	flagConfigNo       = flag.Int("config", -1, "Specify the bootconfiguration to boot. It is the order of the menuentries in grub.cfg")
```
insomniacslk: I think a better default would be `0`. In this way the current behaviour is unchanged, and you don't have to check if the config number is < 0.

nit: maybe a more descriptive name could be `flagConfigIdx`, and the description could become "Specify the index of the configuration to boot. The order is determined by the menu entries in the Grub config"

localboot/main.go
```diff
@@ -123,7 +124,20 @@ func BootGrubMode(devices []storage.BlockDev, baseMountpoint string, guid string
 	if len(bootconfigs) == 0 {
```
insomniacslk: since you're changing this file, can you please also move this `if` above? Just before the log.Printf at line 120. It doesn't make much sense here

localboot/main.go
```diff
@@ -123,7 +124,20 @@ func BootGrubMode(devices []storage.BlockDev, baseMountpoint string, guid string
...
 		return fmt.Errorf("No boot configuration found")
 	}
-
+	for n, cfg := range bootconfigs {
+		log.Printf("  %d: %s\n", n, cfg.Name)
+	}
+	if configNo > -1 {
```
insomniacslk: this changes the current behaviour - from booting the first entry, to trying all the boot entries until one works. This is also different from Grub's behaviour. I would just remove this, and change the default index to 0 as suggested above

insomniacslk (dismissed):

localboot/main.go
jenstopp: I don't understand. The purpose of this flag schould be to choose one specific bootentry from the ones found - which indeet changes the usual behavior. If you don't use the flag the current behavior is applied: trying alle bootconfigs until oneworks

insomniacslk: my bad, scratch that request. We had in store to only try to boot the first entry but we never implemented it, so you're not changing the default behaviour as I initially thought.

(my other requests still apply though)

localboot/main.go
jenstopp: naming sound good, changed it

localboot/main.go
jenstopp: done

localboot/main.go
jenstopp: ok ;)
but I will leave the -1 as default, to detect if the flag is used at all.

zaolin (dismissed):

localboot/main.go
```diff
@@ -116,14 +117,27 @@ func BootGrubMode(devices []storage.BlockDev, baseMountpoint string, guid string
...
 		debug("%+v", cfg)
 	}
-	if len(bootconfigs) == 0 {
-		return fmt.Errorf("No boot configuration found")
+	for n, cfg := range bootconfigs {
+		log.Printf("  %d: %s\n", n, cfg.Name)
+	}
+	if configIdx > -1 {
+		for n, cfg := range bootconfigs {
+			if configIdx == n {
```
insomniacslk, line 132: we should have `dryrun` here too

localboot/main.go
```diff
@@ -116,14 +117,27 @@ func BootGrubMode(devices []storage.BlockDev, baseMountpoint string, guid string
...
+				if err := cfg.Boot(); err != nil {
+					log.Printf("Failed to boot kernel %s: %v", cfg.Kernel, err)
+				}
+			}
+		}
+		log.Printf("Invalig arg -config %d: there are Specify the bootconfigurless then %d bootconfigs available \n", configIdx, len(bootconfigs))
```
insomniacslk: can you fix this log message?

insomniacslk (commented):
two minor changes, otherwise looks good to me

localboot/main.go
jenstopp: done

localboot/main.go
jenstopp: done

PR #106, opened 2019-03-26
Efi integration

jenstopp:
Parse the set-root command from grub file, identify the partition wit the filsystem uuid given in that statatement, then sets the path of kernel, initramfs etc accordingly.
This is necessary since in EFI systems the grub.cfg usually resides on a different partition then the kernel.

For now recognision of ext4 and vfat filesystems is supported

### DIFF
--- a/localboot/grub.go
+++ b/localboot/grub.go
@@ -3,43 +3,54 @@ package main
 import (
 	"io/ioutil"
 	"log"
+	"os"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/systemboot/systemboot/pkg/bootconfig"
 	"github.com/systemboot/systemboot/pkg/crypto"
 )
 
-// List of paths where to look for grub config files. Grub2Paths will look for
-// files with grub2-compatible syntax, GrubLegacyPaths similarly will treat
-// these as grub-legacy config files.
+// List of directories where to look for grub config files. The root dorectory
+// of each mountpoint, these folders inside the mountpoint and all subfolders
+// of these folders are searched
 var (
-	Grub2Paths = []string{
-		// grub2
-		"boot/grub2/grub.cfg",
-		"boot/grub2.cfg",
-		"grub2/grub.cfg",
-		"grub2.cfg",
-	}
-	GrubLegacyPaths = []string{
-		// grub legacy
-		"boot/grub/grub.cfg",
-		"boot/grub.cfg",
-		"grub/grub.cfg",
-		"grub.cfg",
+	GrubSearchDirectories = []string{
+		"boot",
+		"EFI",
+		"efi",
+		"grub",
+		"grub2",
 	}
 )
+
+type grubVersion int
+
+var (
+	grubV1 grubVersion = 1
+	grubV2 grubVersion = 2
+)
+
+func isGrubSearchDir(dirname string) bool {
+	for _, dir := range GrubSearchDirectories {
+		if dirname == dir {
+			return true
+		}
+	}
+	return false
+}
 
 // ParseGrubCfg parses the content of a grub.cfg and returns a list of
 // BootConfig structures, one for each menuentry, in the same order as they
 // appear in grub.cfg. All opened kernel and initrd files are relative to
 // basedir.
-func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.BootConfig {
+func ParseGrubCfg(ver grubVersion, grubcfg string, basedir string) []bootconfig.BootConfig {
 	// This parser sucks. It's not even a parser, it just looks for lines
 	// starting with menuentry, linux or initrd.
 	// TODO use a parser, e.g. https://github.com/alecthomas/participle
-	if grubVersion != 1 && grubVersion != 2 {
-		log.Printf("Warning: invalid GRUB version: %d", grubVersion)
+	if ver != grubV1 && ver != grubV2 {
+		log.Printf("Warning: invalid GRUB version: %d", ver)
 		return nil
 	}
 	bootconfigs := make([]bootconfig.BootConfig, 0)
@@ -65,6 +76,9 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
 			}
 			inMenuEntry = true
 			cfg = new(bootconfig.BootConfig)
+			name := strings.Join(sline[1:], " ")
+			name = strings.Split(name, "--")[0]
+			cfg.Name = name
 		} else if inMenuEntry {
 			// otherwise look for kernel and initramfs configuration
 			if len(sline) < 2 {
@@ -74,17 +88,32 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
 			if sline[0] == "linux" || sline[0] == "linux16" || sline[0] == "linuxefi" {
 				kernel := sline[1]
 				cmdline := strings.Join(sline[2:], " ")
-				if grubVersion == 2 {
-					// if grub2, unquote the string, as directives could be quoted
-					// https://www.gnu.org/software/grub/manual/grub/grub.html#Quoting
-					// TODO unquote everything, not just \$
-					cmdline = strings.Replace(cmdline, `\$`, "$", -1)
-				}
+				cmdline = unquote(ver, cmdline)
 				cfg.Kernel = path.Join(basedir, kernel)
 				cfg.KernelArgs = cmdline
 			} else if sline[0] == "initrd" || sline[0] == "initrd16" || sline[0] == "initrdefi" {
 				initrd := sline[1]
 				cfg.Initramfs = path.Join(basedir, initrd)
+			} else if sline[0] == "multiboot" || sline[0] == "multiboot2" {
+				multiboot := sline[1]
+				cmdline := strings.Join(sline[2:], " ")
+				cmdline = unquote(ver, cmdline)
+				cfg.Multiboot = path.Join(basedir, multiboot)
+				cfg.MultibootArgs = cmdline
+			} else if sline[0] == "module" || sline[0] == "module2" {
+				module := sline[1]
+				cmdline := strings.Join(sline[2:], " ")
+				if ver == grubV2 {
+					// if grub2, unquote the string, as directives could be quoted
+					// https://www.gnu.org/software/grub/manual/grub/grub.html#Quoting
+					// TODO unquote everything, not just \$
+					cmdline = strings.Replace(cmdline, `\$`, "$", -1)
+				}
+				module = path.Join(basedir, module)
+				if cmdline != "" {
+					module = module + " " + cmdline
+				}
+				cfg.Modules = append(cfg.Modules, module)
 			}
 		}
 	}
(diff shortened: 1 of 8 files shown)


jenstopp:
depends on #91

localboot/grub.go
```diff
@@ -65,7 +86,43 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
 			}
 			inMenuEntry = true
 			cfg = new(bootconfig.BootConfig)
+			name := ""
+			if len(sline) > 1 {
+				name = strings.Join(sline[1:], " ")
+				name = unquote(ver, name)
+				name = strings.Split(name, "--")[0]
+			}
+			cfg.Name = name
```
tfg13, line 98: I see this was discussed already in the other PR. It can produce ugly names, but is OK for now, since we need to unify the grub parsers soon anyway.

localboot/grub.go
```diff
@@ -65,7 +86,43 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
 		} else if inMenuEntry {
+			// check if location of kernel is at an other partition
+			// see https://www.gnu.org/software/grub/manual/grub/html_node/search.html
+			if sline[0] == "search" {
+				for _, str1 := range sline {
+					if str1 == "--set=root" {
+						log.Printf("Kernel seems to be on an other partitioin then the grub.cfg file")
+						for _, str2 := range sline {
+							if len(str2) == 36 && string(str2[8]) == "-" && string(str2[13]) == "-" && string(str2[18]) == "-" && string(str2[23]) == "-" {
```
tfg13: I rarely suggest regexes as solutions, but this here cries for one :D

localboot/grub.go
```diff
@@ -65,7 +86,43 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+								log.Printf("fs-uuid: %s", kernelFsUUID)
+								partitions, err := storage.PartitionsByFsUUID(devices, kernelFsUUID)
+								if err != nil {
+									log.Printf("Unexpected error while looking up fs uuid: %v", err) // PartitionsByFsUUID does not return an error for now
+								} else if len(partitions) == 0 {
+									log.Printf("WARNING: No partition found with filesystem uuid:'%s' to load kernel from!", kernelFsUUID) // TODO throw error ?
+									continue
+								}
+								if len(partitions) > 1 {
+									log.Printf("WARNING: more than one partition found with the given file. Using the first one")
```
tfg13:
```suggestion
									log.Printf("WARNING: more than one partition found with the given fs UUID. Using the first one")
```

localboot/grub.go
```diff
@@ -65,7 +86,43 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
+								}
+								dev := partitions[0]
+								kernelBasedir = path.Dir(kernelBasedir)
+								kernelBasedir = path.Join(kernelBasedir, dev.Name)
```
tfg13, line 120: I think you have to reset `kernelBasedir` back to `basedir` after this menuentry is over. Otherwise, if there is an entry after this, that does not have `search`, it will try to find the kernel on this disk. Or can this never happen?

pkg/bootconfig/bootconfig.go
```diff
@@ -3,69 +3,97 @@ package bootconfig
...
-	crypto.TryMeasureBootConfig(bc.Name, bc.Kernel, bc.Initramfs, bc.KernelArgs, bc.DeviceTree)
-
-	kernel, err := os.Open(bc.Kernel)
-	if err != nil {
-		return err
-	}
-	var initramfs *os.File
-	if bc.Initramfs != "" {
-		initramfs, err = os.Open(bc.Initramfs)
+	data := bc.Name + bc.Kernel + bc.Initramfs + bc.KernelArgs + bc.DeviceTree + bc.Multiboot + bc.MultibootArgs
```
tfg13: Picking up where the previous review left off: This should come out of some function we call on `BootConfig`. Then that config is responsible to include all the things we want to measure. (`Modules` is still missing, I think)

pkg/bootconfig/bootconfig.go
```diff
@@ -3,69 +3,97 @@ package bootconfig
...
+		// }
+		// trampoline, err := filepath.EvalSymlinks(p)
+		// if err != nil {
+		// 	return fmt.Errorf("Cannot eval symlinks for %v: %v", p, err)
+		// }
+		// load multiboot kernel and modules
+		// m := multiboot.New(bc.Multiboot, bc.MultibootArgs, trampoline, bc.Modules)
+		// if err := m.Load(true); err != nil {
+		// 	return fmt.Errorf("Load failed: %v", err)
+		// }
```
tfg13: What is the story with this commented block, can it go?

pkg/storage/blockdev.go
```diff
@@ -216,6 +370,18 @@ func PartitionsByGUID(devices []BlockDev, guid string) ([]BlockDev, error) {
 	return partitions, nil
 }
 
+// PartitionsByFsUUID returns a list of BlockDev objects whose underlying
+// block device has a filesystem with the given UUID
+func PartitionsByFsUUID(devices []BlockDev, fsuuid string) ([]BlockDev, error) {
```
tfg13: Why does this return an error if it never happens?

pkg/bootconfig/bootconfig.go
```diff
@@ -3,69 +3,97 @@ package bootconfig
...
-
-	kernel, err := os.Open(bc.Kernel)
-	if err != nil {
-		return err
-	}
-	var initramfs *os.File
-	if bc.Initramfs != "" {
-		initramfs, err = os.Open(bc.Initramfs)
+	data := bc.Name + bc.Kernel + bc.Initramfs + bc.KernelArgs + bc.DeviceTree + bc.Multiboot + bc.MultibootArgs
+	crypto.TryMeasureData(crypto.BootConfigPCR, []byte(data), "bootconfig")
```
tfg13: Doesn't this produce a different result than previously with `TryMeasureBootConfig` calling `Measure` separately on each string?

pkg/crypto/measure.go
```diff
@@ -8,29 +8,34 @@ import (
...
+	BootConfigPCR uint32 = 8
+	// ConfigDataPCR type in PCR 8
+	ConfigDataPCR uint32 = 8
+	// NvramVarsPCR type in PCR 9
+	NvramVarsPCR uint32 = 9
 )
 
 // TryMeasureBootConfig measures bootconfig contents
-func TryMeasureBootConfig(name, kernel, initramfs, kernelArgs, deviceTree string) {
+func TryMeasureBootConfig(name, kernel, initramfs, kernelArgs, deviceTree, multiboot, multibootArgs string, modules []string) {
```
tfg13: Is this function still used now that `Boot()` does not call it anymore?

pkg/storage/blockdev.go
```diff
@@ -165,11 +167,163 @@ func GetBlockStats() ([]BlockDev, error) {
...
+// 	} else {
+// 		fields := strings.Fields(string(output))
+// 		for _, field := range fields {
+// 			if strings.Contains(field, "UUID=") && !strings.Contains(field, "PARTUUID=") {
+// 				fsuuid = strings.Trim(field, `UUID"=`)
+// 			}
+// 		}
+// 		return fsuuid, nil
+// 	}
+// }
```
tfg13: I am assuming this can go?

localboot/grub.go
tfg13: Wait, isn't the UUID of a `VFAT` shorter?

pkg/storage/blockdev.go
```diff
@@ -165,11 +167,163 @@ func GetBlockStats() ([]BlockDev, error) {
...
+
+func getUUID(devpath string) (fsuuid string) {
```
tfg13, line 177: We can keep this for now, but I wonder if we could avoid hardcoding all this filesytem specific stuff by just reading everything in `/dev/disk/by-uuid` once and then doing a reverse matching with the mount point.

The main reason why I started to think about this is because I realized we also need `btrfs` here :D

tfg13 (changes requested):
Some minor changes are still required. We should be able to finish this today :)

localboot/grub.go
tfg13: Huh? Not sure how this came here

tfg13:
Ah and it does not compile in travis, I think this is because of the restructuring of u-root, you might have to rebase u-root (again)

localboot/grub.go
jenstopp: yes, you are right. But I think it is OK for now, too. Will be done better when the grub parser is there.

pkg/bootconfig/bootconfig.go
jenstopp: I have reworked the measurement of the bootconfigs

localboot/grub.go
jenstopp: Yes, it has to be reset after each menuentry. Done.

pkg/storage/blockdev.go
jenstopp: Done

pkg/crypto/measure.go
jenstopp: No it is not used. Removed it.

pkg/bootconfig/bootconfig.go
jenstopp: removed it

pkg/bootconfig/bootconfig.go
jenstopp: Now it is measured in the same way as it was intended by MeasureBootConfig()

pkg/storage/blockdev.go
jenstopp: dev/disk/by-uuid is not available

pkg/storage/blockdev.go
jenstopp: I have removed the error

pkg/storage/blockdev.go
tfg13: Awut? I think it should be :D

Anyway, this was more of a general comment for the future, we can keep this for now. But we will need `btrfs` support eventually.

localboot/grub.go
tfg13: We only support booting by UUIDs for now, right? I think we can leave this as-is for now, but we need to revamp this with support for labels.

localboot/grub.go
```diff
@@ -74,56 +132,113 @@ func ParseGrubCfg(grubcfg string, basedir string, grubVersion int) []bootconfig.
...
 	}
+
 	// append last kernel config if it wasn't already
 	if inMenuEntry && cfg.IsValid() {
 		bootconfigs = append(bootconfigs, *cfg)
 	}
 	return bootconfigs
 }
 
+func isValidFsUUID(uuid string) bool {
```
tfg13, line 167: This will probably return true for lots of things we don't regard as valid UUIDs, for example it will fail to parse:
`search --set=root --label abcd`

Then again, booting with labels is not support right now anyway, and we are going to revamp the grub parser, this is fine for now.

tfg13 (approved):
We can integrate this now. I will file some issues as soon as the move to u-root is done against things like revamping the grub parser(s), supporting labels, and either adding support for `btrfs` UUIDs or revamping it altogether.

PR #112, opened 2019-04-15
Don't read CA Certs from VPD, read them from the filesystem instead

BigValen:
Test Plan: Boot host with these settings, with image built with -files=cacerts.pem:etc/cacerts.pem


insomniacslk (approved):
lgtm!

PR #120, opened 2019-05-03
[netboot] Call fixmynetboot if netboot fails

insomniacslk:
fixmynetboot is now part of the netboot workflow. It is called
automatically if netboot fails, if speficied in the VPD variables,
or if `-fix` is passed.
    
Also corrected error handling logic in checker and restructured command
execution.

NOTE: this changes the default behaviour of `netboot`. To revert that, uinit should call `netboot -nofix`. However this should be safe, as `fixmynetboot` does non-destructive, fast checks.

Signed-off-by: Andrea Barberio <insomniac@slackware.it>

### DIFF
--- a/cmds/fixmynetboot/main.go
+++ b/cmds/fixmynetboot/main.go
@@ -1,22 +1,105 @@
 package main
 
 import (
+	"flag"
+	"fmt"
+	"log"
+	"net"
+
 	"github.com/systemboot/systemboot/pkg/checker"
 )
 
 // fixmynetboot is a troubleshooting tool that can help you identify issues that
 // won't let your system boot over the network.
+// NOTE: this is a DEMO tool. It's here only to show you how to write your own
+// checks and remediations. Don't use it in production.
 
-func main() {
+var emergencyShellBanner = `
+**************************************************************************
+** Interface checks failed, see the output above to debug the issue.     *
+** Entering the emergency shell, where you can run "fixmynetboot" again, *
+** or any other LinuxBoot command.                                       *
+**************************************************************************
+`
+
+var (
+	doEmergencyShell = flag.Bool("shell", false, "Run emergency shell if checks fail")
+)
+
+func checkInterface(ifname string) error {
 	checklist := []checker.Check{
-		checker.Check{"wlp2s0 exists", checker.InterfaceExists("wlp2s0"), nil, false},
-		checker.Check{"eth0 exists", checker.InterfaceExists("eth0"), checker.InterfaceRemediate("eth0"), false},
-		checker.Check{"eth0 link speed", checker.LinkSpeed("eth0", 100), nil, false},
-		checker.Check{"eth0 link autoneg", checker.LinkAutoneg("eth0", true), nil, false},
-		checker.Check{"eth0 has link-local", checker.InterfaceHasLinkLocalAddress("wlp2s0"), nil, false},
-		checker.Check{"eth0 has global addresses", checker.InterfaceHasGlobalAddresses("wlp2s0"), nil, false},
+		checker.Check{
+			Name:        fmt.Sprintf("%s exists", ifname),
+			Run:         checker.InterfaceExists(ifname),
+			Remediate:   checker.InterfaceRemediate(ifname),
+			StopOnError: true,
+		},
+		checker.Check{
+			Name:        fmt.Sprintf("%s link speed", ifname),
+			Run:         checker.LinkSpeed(ifname, 100),
+			Remediate:   nil,
+			StopOnError: false},
+		checker.Check{
+			Name:        fmt.Sprintf("%s link autoneg", ifname),
+			Run:         checker.LinkAutoneg(ifname, true),
+			Remediate:   nil,
+			StopOnError: false,
+		},
+		checker.Check{
+			Name:        fmt.Sprintf("%s has link-local", ifname),
+			Run:         checker.InterfaceHasLinkLocalAddress(ifname),
+			Remediate:   nil,
+			StopOnError: true,
+		},
+		checker.Check{
+			Name:        fmt.Sprintf("%s has global addresses", ifname),
+			Run:         checker.InterfaceHasGlobalAddresses("eth0"),
+			Remediate:   nil,
+			StopOnError: true,
+		},
 	}
 
-	checker.Run(checklist)
+	return checker.Run(checklist)
+}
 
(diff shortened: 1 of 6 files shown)


insomniacslk:
This is ready to review

cmds/fixmynetboot/main.go
```diff
@@ -1,22 +1,44 @@
...
 func main() {
+	flag.Parse()
 	checklist := []checker.Check{
-		checker.Check{"wlp2s0 exists", checker.InterfaceExists("wlp2s0"), nil, false},
-		checker.Check{"eth0 exists", checker.InterfaceExists("eth0"), checker.InterfaceRemediate("eth0"), false},
+		checker.Check{"eth0 exists", checker.InterfaceExists("eth0"), checker.InterfaceRemediate("eth0"), true},
 		checker.Check{"eth0 link speed", checker.LinkSpeed("eth0", 100), nil, false},
 		checker.Check{"eth0 link autoneg", checker.LinkAutoneg("eth0", true), nil, false},
-		checker.Check{"eth0 has link-local", checker.InterfaceHasLinkLocalAddress("wlp2s0"), nil, false},
+		checker.Check{"eth0 has link-local", checker.InterfaceHasLinkLocalAddress("wlp2s0"), nil, true},
```
pmazzini: Is this eth0 or wlp2s0?

pmazzini:
> It is called automatically if netboot fails

Do we actually want this? What about the case where a server is set to boot from disk and the dhcp server doesn't reply with a boot url?

Will it be called every time we set a machine to boot from disk?

cmds/fixmynetboot/main.go
insomniacslk: good catch, thanks

insomniacslk:
> > It is called automatically if netboot fails
> 
> Do we actually want this? What about the case where a server is set to boot from disk and the dhcp server doesn't reply with a boot url?

I am still refining the logic here, but this will be controlled by boot variables, and even in that case you still need to explicitly include `fixmynetboot` in the ramdisk for this to happen.

Additionally, this is a sample CLI for demo purposes, but the actual desired behaviour will be codified via JSON configuration when the `checker` command will be implemented (currently in development).

> Will it be called every time we set a machine to boot from disk?

Only if that machine tries netboot, and if netboot fails. If the system is configured to boot straight from disk, netboot won't even be tried.

On systems where netboot is always the first tentative (and disk booting the fallback), the NIC troubleshooting is instantaneous and non-disruptive. I agree that the netboot URL check should be removed. I will remove the DHCP netboot check.

pmazzini:
> Only if that machine tries netboot, and if netboot fails. If the system is configured to boot straight from disk, netboot won't even be tried.

I am still thinking about our particular use case where we do not change boot variables but control the dhcp server instead to decide whether a server should netboot or not.

I wouldn't make the default behavior to call `fixmynetboot` upon failing to netboot.

insomniacslk:
> > Only if that machine tries netboot, and if netboot fails. If the system is configured to boot straight from disk, netboot won't even be tried.
> 
> I am still thinking about our particular use case where we do not change boot variables but control the dhcp server instead to decide whether a server should netboot or not.
> 
> I wouldn't make the default behavior to call `fixmynetboot` upon failing to netboot.

That's the use case that I described in the last sentence of my reply. Calling `fixmynetboot` with the demo configuration will only print diagnostic output. It won't slow the boot down or make disruptive changes. In my opinion this is desirable in any case

netboot/main.go
```diff
@@ -108,6 +110,16 @@ func main() {
 		}
 		for _, d := range dhcp {
 			if err := boot(iface.Name, d); err != nil {
+				if !*noFixmynetboot {
+					cmd := exec.Command("fixmynetboot")
```
pmazzini: By default I think we loop over all available interfaces. However, `fixmynetboot` only checks `eth0`. So this may call `fixmynetboot` several times for `eth0`.

insomniacslk: Good catch! I've updated the code to run fixmynetboot only on the failing interface

cmds/fixmynetboot/main.go
```diff
@@ -1,22 +1,105 @@
...
+	} else {
+		interfaces = []string{ifname}
+	}
 
+	for _, ifname := range interfaces {
+		if err := checkInterface(ifname); err != nil {
+			if !*doEmergencyShell {
+				log.Fatal(err)
+			}
+			if err := checker.EmergencyShell(dhcpErrorBanner)(); err != nil {
```
pmazzini: Is this dhcp banner still relevant?

netboot/main.go
```diff
@@ -108,6 +110,16 @@ func main() {
 		}
 		for _, d := range dhcp {
 			if err := boot(iface.Name, d); err != nil {
+				if !*noFixmynetboot {
+					cmd := exec.Command("fixmynetboot", "-shell", iface.Name)
```
pmazzini: Does it drops into an emergency shell if one of the interfaces fails a check?

cmds/fixmynetboot/main.go
insomniacslk: Not anymore, I'll leave it out

netboot/main.go
insomniacslk: Yes, good catch again. I'll drop the `-shell` parameter.
Thing is - I'm preparing a demo where I need the DHCP check plus others that should not be the default, and some demo code is left in. Sorry about that.

cmds/fixmynetboot/main.go
insomniacslk: actually I'll just change the message to be generic rather than about DHCP

pmazzini: Should the `dhcp` name also be dropped?

insomniacslk:
I've changed the code to minimize the behaviour change. The default  is again to just do netboot without calling fixmynetboot. The latter is only triggered if specified via the newly added DebugOnFailure option in the netboot Booter, or if `-fix` is passed explicitly.
I still need to trigger an emergency shell. Thinking of adding a `ShellOnFailure` field to the booter, but at the same time I don't want to add too many flags.

cmds/fixmynetboot/main.go
insomniacslk: done

pmazzini (approved):
> I've changed the code to minimize the behaviour change. The default is again to just do netboot without calling fixmynetboot.

Ok, makes more sense to me.

PR #121, opened 2019-05-10
[checker] Checker bin, JSON config file, new Check interface

kawmarco:
* Added checker command line entry point (`checker/main.go`) which accepts a JSON configuration file as input (`--config-file`), containing a list of `Check`s
 * Adapted `Check` structs to be more friendly to JSON configuration and JSON report output, thinking of the following workflow
    * Developers define `CheckFun(CheckArgs)` functions, which contain the actual logic of the check, and "register" their functions using `registerCheckFun` on `init()` (so that they are registered in a global map and we can then find the check functions by name at runtime)
    * A JSON-friendly `Check` struct defines how `CheckFuns` should be called, including `CheckArgs`, possible remediations and metadata. The config file is a `checklist` (a list of `Check`)
    * Calling `Check.Run()` returns a JSON-friendly `CheckResult`,  containing the result of the `Check` and the result of the configured remediations

TODO:
 * [x] Document `pkg/checker/checker.go`
 * [x] Proper error/exit value on invalid/insufficient arguments (maybe just support one positional argument)?
 * [x] Ensure checker binary is in the ramdisk


insomniacslk:
hey @kawmarco , thanks for the contribution! Could you please re-submit with DCO (just amend the commit with `-s`), and address the linter warnings?
You may want to run `golangci-lint run` and fix any other warning before re-submitting (it will be caught by the CI anyway)

README.md
```diff
@@ -58,8 +58,9 @@ The `uinit` program just wraps `netboot` and `localboot` in a forever-loop logic
 
 ```
 go get -u github.com/u-root/u-root
-go get -u github.com/systemboot/systemboot/{uinit,localboot,netboot}
-u-root -build=bb core github.com/systemboot/systemboot/{uinit,localboot,netboot}
+go get -u github.com/safchain/ethtool
+go get -u github.com/systemboot/systemboot/{uinit,localboot,netboot,checker}
+u-root -build=bb core github.com/systemboot/systemboot/{uinit,localboot,netboot,checker}
```
insomniacslk: I don't think we should immediately recommend `checker` at this stage, I'd let it bake in and get some larger scale testing before calling it stable

checker/main.go
```diff
@@ -0,0 +1,66 @@
+package main
```
insomniacslk: this should go under `cmds/checker`, we are moving to a structure where `pkg` contains libraries and `cmds` contains programs, and nothing else at the top level

checker/main.go
```diff
@@ -0,0 +1,66 @@
...
+	"io/ioutil"
+	"log"
+	"os"
+
+	"github.com/systemboot/systemboot/pkg/checker"
+)
+
+var configFile string
+
+func init(){
```
insomniacslk: nit: I have a strong preference for minimizing the use of `init` functions (because of how they could be used to introduce implicit behaviours). In this case it's not even necessary - you can just declare it as
```
var configFile = flag.String("c", "", "Configuration file that defines checks and remediations")
```

Note that I also suggested to use "-c" instead of "-config-file" so we minimize the CLI verbosity

checker/main.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func usage(){
+	fmt.Fprintf(os.Stderr, "Usage of %s:\n", os.Args[0])
+	flag.PrintDefaults()
+}
+
+func run() int {
+	if configFile == "" {
+		fmt.Fprintf(os.Stderr, "Error: config-file argument is required\n")
```
insomniacslk: nit: better to use `CommandLine.Output()` instead of `os.Stderr`, this allows overriding where usage is printed (with `SetOutput`)

checker/main.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/systemboot/systemboot/pkg/checker"
+)
+
+var configFile string
+
+func init(){
+	flag.StringVar(&configFile, "config-file", "", "Config file to use")
+}
+
+func usage(){
```
insomniacslk: this is not necessary, you can call just `flag.Usage()`

checker/main.go
```diff
@@ -0,0 +1,66 @@
...
+	if numErrors > 0 {
+		return 1
+	}
+
+	return 0
+}
+
+func main() {
+	flag.Parse()
+	ret := run()
```
insomniacslk: why splitting in `main` and `run`? We don't gain in readability nor anything else (unless I'm missing something)

checker/main.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func run() int {
+	if configFile == "" {
+		fmt.Fprintf(os.Stderr, "Error: config-file argument is required\n")
+		usage()
+		return 1
+	}
+
+	log.Printf("Registered Checks: %v\n", checker.ListRegistered())
```
insomniacslk: I think this should go behind a flag, e.g. `-d` for debug

checker/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+	var checklist []checker.Check
+
+	checkerConfigStr, err := ioutil.ReadFile(configFile)
+	if err != nil {
+		log.Printf("Unable to open config file %#v: %s", configFile, err.Error())
```
insomniacslk: you should invert the format strings - use `%s` for `configFile` and `%v` for `err`. Also just use `err`:
```suggestion
		log.Printf("Unable to open config file %s: %v", configFile, err)
```
(note: when moving `configFile` outside `init`, you will need to dereference it here)

checker/main.go
```diff
@@ -0,0 +1,66 @@
...
+		return 1
+	}
+
+	err = json.Unmarshal(checkerConfigStr, &checklist)
+	if err != nil {
```
insomniacslk: just do
```
if err := json.Unmarshal(...); err != nil {
```

checker/main.go
```diff
@@ -0,0 +1,66 @@
...
+		log.Printf("Unable to parse config file %#v: %s\n", configFile, err.Error())
+		return 1
+	}
+
+	checklistJSON, _ := json.MarshalIndent(checklist, "", "    ")
```
insomniacslk: even if you don't expect an error, we should check for it anyway.

This should be printed only when running in debug mode

checker/main.go
```diff
@@ -0,0 +1,66 @@
...
+	log.Printf("Checklist: %s\n", checklistJSON)
+
+	results, numErrors := checker.Run(checklist)
+	resultsJSON, _ := json.MarshalIndent(results, "", "    ")
```
insomniacslk: same here

pkg/checker/checker.go
```diff
@@ -2,49 +2,113 @@ package checker
...
+
+	Remediations  []Check `json:"remediations"`
+	StopOnFailure bool    `json:"stopOnFailure"`
+}
+
+// A CheckResult describes the result of Running a Check
+type CheckResult struct {
+	Description string `json:"description"`
+
+	CheckFunName string    `json:"checkFun"`
```
insomniacslk: I guess we should call this `CheckerFuncName` ("func" instead of "fun") for consistency and clarity

insomniacslk (changes requested):
I've done only a partial review, publishing it to unblock you.

Also note - several files have changed permissions with this commit, they have now an executable bit set where they should not

checker/main.go
kawmarco: The idea was to make `run()` less side-effecty and easier to unit test (e.g. by returning an `int` instead of calling `os.Exit()`)

tfg13:
Discussed offline, we will reopen this one against uroot after the move is done

